PR #162, opened 2022-07-03
External host setting

### LINKED ISSUE #158: Show actual host and port in Target popup
It would be nice if this popup could show the actual SSH command that can be used:

image

Eugeny:
Fixes #158


bram-pkg:
Damn man, you're on a roll!

PR #315, opened 2022-09-13
added support for user role management

### LINKED ISSUE #314: Manage user roles in webui
Hey,

i have followed this project so long and just waited for the features in release 0.6.0. Awesome!
There is just a single feature missing:
You can't manage the roles of a user in the web ui at least i still haven't found it. I think this is mandatory for version 0.6.0 because you already can manage the roles for a target.

Keep going!

Gurkengewuerz:
well this was easier than i thought because you already have implemented the full api. Fixes #314
i would appreciate the merge and also if you could trigger a docker image upgrade very soon :)

image


Eugeny:
Great work @Gurkengewuerz and yes, I've totally missed this :D

Eugeny:
@all-contributors please add @Gurkengewuerz  for code

Issue #440, opened 2022-10-27
Forward login name in ssh connection


notnooblord:
Currently in ssh target we manually specify username, or can use blank.

But blank username does just what it means, it sets no login name for ssh, so results in `sshd[3301684]: Invalid user`

Would be great if blank option just forwarded login name used to authenticate to bastion, can this be done?

notnooblord:
I have currently just patched [here](https://github.com/warp-tech/warpgate/blob/main/warpgate-protocol-ssh/src/server/session.rs#L1480): 
```rust
if ssh_options.username.is_empty() {
    ssh_options.username = username.to_string();
}
```
This is really needed as it allows users to login under their username, one thing missing is forwarded keys - hope it gets implemented :)
I wonder if there is some better solution tho. I'm willing to do PR for this.

Eugeny:
Happy to accept the PR if you also add an explanatory placeholder for the [username field](https://github.com/warp-tech/warpgate/blob/main/warpgate-web/src/admin/Target.svelte#L150-L152)!

notnooblord:
Filled in quick PR #445, I also stumbled upon a bug with placeholder text:

https://user-images.githubusercontent.com/11678665/198910841-38972310-8d6c-4d2c-b727-0e3f28fb621d.mov

Should be fixed in PR (at least for me it works OK).

Please let me now if any thing else to be done, and happy to contribute!

PR #445, opened 2022-10-31
Feature - forward username

notnooblord:
Relates: #440

If username is not present in ssh config - we take in from current authenticated user and try to login with it.  
Since there is no point in blank username field, added some placeholder text to it, explaining username forwarding.  

Also there is some bug with floating FormGroup placeholder text in bootstrap 5, this simple css fix helped: https://github.com/twbs/bootstrap/issues/33999#issuecomment-843540381.


Eugeny:
Thanks!

Issue #453, opened 2022-11-02
Where in TIMEOUT=120 poetry run ./run.sh does it fail


stappersg:
Hi,

This issue is for tracking some _work in progress_ related to #446 

The plan is to have this closed when all the _Bump foo version to version+n_ backlog is gone.

stappersg:
Since https://github.com/warp-tech/warpgate/issues/446#issuecomment-1300991838 no need to keep this tracking issue open.

Issue #446, opened 2022-10-31
Merge request back log, help offer.


stappersg:
Hello Warp-tech people,

Currently there are 36 pending merge requests.  Most of them are from @dependabot.

I'm offering help to merge those.

Benefit for the project:  No more back log that screams _this project is abonded_

My benefit: Contributing to the success of this project.

Regards
Geert Stappers
My email addresses is `stappers@stappers.it` as documented in public visible part of my Github profile.

Eugeny:
Hey, I normally juts sort through these once a month - currently working on expanding the test suite before merging more. But it'd be awesome if you could go through these and merge relevant ones (based on release notes) into a single PR! (dependabot will automatically close individual PRs once it's merged)

Eugeny:
This error is going to be caused by some library referencing v2.18+ version symbols in glibc. The builds are running against glibc 2.17 to retain compatibility with RHEL 7 - I would like to keep it that way unless there's some clear benefit.

Eugeny:
Also something to keep in mind - there's a flaky test (`test_ssh_proto.py::Test::test_direct_tcpip`) that almost always fails in CI but doesn't locally - I haven't had time to debug it yet - but you can safely ignore it.

stappersg:
> almost always fails in CI ... but you can safely ignore it

I can and will ignore that failing test.

However, I think that _dependabot_ doesn't do so,  the reason for the merge request backlog ...

Eugeny:
Dependabot doesn't merge anything automatically - it just creates PRs.

stappersg:
> Dependabot just creates PRs

I would like to have the privilege to approve merge request. Benefit for the _warpgate project_ is (much) less backlog on MRs from @dependabot.

Eugeny:
For starters, please just cherry-pick dependabot's commits into a single PR of your own - this way I'll have a chance to review it before merging.

stappersg:
> >   blunt github merge

> For starters, please just cherry-pick dependabot's commits into a single PR of your own - this way I'll have a chance to review it before merging.

I think that the idea is avoid those ugly commit messages that github puts in. I like that idea and have started working on it.

For reviewing is this my advice
```text
git remote add stappers https://git.sr.ht/~stappers/warpgate
git fetch stappers
git switch dependabotcherries
git log --oneline | head -n 15
git log --patch
ctrl-C on reaching Bump russh from 0.34.0-beta.16 to 0.34.0-beta.17
```

[...]

PR #454, opened 2022-11-02
Added debug print statements to tests/run.sh

stappersg:
It is the `set -x` that is the important.

The _print working direcoty_ are for getting information for running the test outside the CI.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -e
+set -x
+pwd
 cd ..
+pwd
 rm target/llvm-cov-target/* || true
 cargo llvm-cov clean --workspace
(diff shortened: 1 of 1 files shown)


Eugeny:
Could you elaborate a bit? The script is designed to be run from the `tests` directory, and you can just run it with `bash -x run.sh`

stappersg:
It is for #453.

And might **not** be needed to be merged, because making a _merge request_ does trigger `CI`.

stappersg:
(Our comments did cross mid air.)

> Could you elaborate a bit?

#453 links to #446.  Solving #446 is the real goal.

stappersg:
https://github.com/warp-tech/warpgate/issues/446#issuecomment-1300991838 says it is a known issue.

PR #828, opened 2023-07-07
Add linux/arm64 docker build

zasdaym:
See title.


Eugeny:
Thanks for contributing and sorry for the delay! @all-contributors please add @zasdaym for code

Eugeny:
I've had to revert this unfortunately since the builds became flaky. Would appreciate if you could take a second look :v:

zasdaym:
Will do. I think the Dockerfile also need to be adjusted.

Issue #885, opened 2023-08-31
Incorrect SSO URL with reverse proxy


hdhog:
I use traefik as reverse proxy for warpgate

traefik pass headers `Host` `X-Forwarded-Proto` to warpgate. For test headers i replace warpgate to [whoami](github.com/traefik/whoami) and see this.

```
Hostname: b3b6d276c624
IP: 127.0.0.1
IP: 172.17.0.2
RemoteAddr: 172.16.11.10:39754
GET / HTTP/1.1
Host: wg.host.me
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/116.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.5
Cookie: ***
Dnt: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
Sec-Gpc: 1
Te: trailers
Uber-Trace-Id: 6f7bcaeb7c5d46db:38dadea1baa77b35:27ac223ed317cf47:1
Upgrade-Insecure-Requests: 1
X-Forwarded-For: ****
X-Forwarded-Host: wg.host.me
X-Forwarded-Port: 443
X-Forwarded-Proto: https
X-Forwarded-Server: 6681ce2d84dd
X-Real-Ip: ***
```

request contains correct headers

but redirect URL created with port 8888 `https://wg.host.me:8888/@warpgate/api/sso/return?code=*&scope=openid+email&state=*` instead `https://wg.host.me/@warpgate/api/sso/return?code=*&scope=openid+email&state=*`

hdhog:
hm. interesting.

i added this  code to `construct_external_url` 
```rust
for x in for_request {
    for x in x.headers() {
        print!("{}={:?}\n",x.0,x.1);
    }
}
```

and don't see `Host` header

```
авг 31 20:50:59 srv warpgate[3061522]: referer="https://wg.host.me/@warpgate"
авг 31 20:50:59 srv warpgate[3061522]: x-real-ip="172.16.11.254"
авг 31 20:50:59 srv warpgate[3061522]: sec-fetch-dest="empty"
авг 31 20:50:59 srv warpgate[3061522]: te="trailers"
[...]

PR #886, opened 2023-09-04
feat:  added external_proxy_header option 

hdhog:
#885 added new options `external_proxy_header` 

defaults
```
external_proxy_header:
  host: "host"
  proto: "x-forwarded-proto"
```

### DIFF
--- a/warpgate-common/src/config/defaults.rs
+++ b/warpgate-common/src/config/defaults.rs
@@ -71,11 +71,11 @@ pub(crate) fn _default_ssh_keys_path() -> String {
 }
 
 #[inline]
-pub(crate) fn _default_external_proxy_header_host_header() -> String {
+pub(crate) fn _default_external_proxy_header_host() -> String {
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you for the PR! Is your plan to use `host: x-forwarded-host` in your configuration? In this case, do you think it would make more sense to simply add a `trust_proxy_headers` config option and allow fallback from `x-forwarded-host` to the normal `host` so that it works both "internally" and "externally"?

lanrat:
@Eugeny IS there any chance you could merge this or another fix for the host header behind a reverse proxy?

Currently, Warpgate is broken when behind a reverse proxy due to it always adding its internal listen url to the callback.

Skyler84:
I've had Warpgate working fine behind a reverse proxy (Traefik), as long as the `external_host` field is set warpgate redirects correctly!

Eugeny:
Closing in favour of  #907

PR #896, opened 2023-09-26
Fix redirection with a relative location

### LINKED ISSUE #891: Relative redirect location returns an "invalid format" error
Hello,

I played with warpgate for some times and encountered a case where an HTTP target don't redirect correctly when the location is not starting with a slash. For exemple, I have a case where target app have an endpoint `/v2` that returns a 301 with a header `location: v2/` and my browser correctly redirects to `/v2/`. I don't know if the RFC accepts that URI format, or if my browser is just cool about it

I tried to figure which part fail to parse this location header, and I still don't know if it's either warpgate, poem or reqwest. I also don't know if the fix be in the on one of the module listed before, or the target application if the format is not standard.

This bug is not fatal since I can manually add the slash to make it works.

seysn:
Fixes #891


Eugeny:
Nice work! 💪  @all-contributors please add @seysn for code

PR #920, opened 2023-11-10
Add support for role mappings on custom SSO providers.

Skyler84:
This is done using the `role_mappings` property.
Roles to be mapped are gotten from the 'warp_groups` oidc claim:
```sso_providers: 
  - name: custom_sso
    label: Custom SSO
    provider:
      type: custom
      client_id: <client_id>
      client_secret: <client_secret>
      issuer_url: <issuer_url>
      scopes: ["email", "profile", "openid", "warp_groups"] #warp_groups is scope name to request for my demo case, which adds a "warpgate_groups" claim to the userinfo
      role_mappings: 
        - ["warpgate:admin", "warpgate:admin"]
```
This maps the `warpgate:admin` group from OIDC to the `warpgate:admin` role.

This [video on YouTube](https://youtu.be/XCYSGGCgk9Q) demonstrates the functionality

PR #921, opened 2023-11-13
Add support for trusting X-Forwarded-For header to get client IP

### LINKED ISSUE #882: Accept X-Forwarded-For from a proxy in front of Warpgate
Hello, 
i tried to run warpgate behind haproxy but i still have the ip address of my haproxy in my logs. is there any configuration to do to have the public ip address?
backend haproxy : 

backend http_back_gate
   balance roundrobin
   option httpclose
   cookie JSESSIONID prefix
   option forwardfor
   http-request set-header X-Forwarded-Proto https if { ssl_fc }
   http-request set-header X-Forwarded-Proto http if !{ ssl_fc }
   http-request redirect scheme https unless { ssl_fc }
   server server1 IP:8888 check ssl ca-file root.pem verify required
   
   
CleanShot 2023-08-28 at 11 07 02

Skyler84:
This fixes #882 
falls back to remote ip if header unavailable

Added `trust_x_forwarded_headers` option in warpgate.yaml config
```yaml
http:
  enable: true
  listen: "0.0.0.0:8888"
  certificate: "./data/tls.certificate.pem"
  key: "./data/tls.key.pem"
  trust_x_forwarded_headers: false
```

### DIFF
--- a/warpgate-common/src/config/mod.rs
+++ b/warpgate-common/src/config/mod.rs
@@ -303,7 +303,11 @@ impl WarpgateConfig {
         // if trust x-forwarded, get x-forwarded-proto, then try request scheme, then fallback https
         // if trust x-forwarded, get x-forwarded-port, then try request port, then fallback http listen port
         let trust_forwarded_headers = self.store.http.trust_x_forwarded_headers;
-        let (scheme, host, port) = ("https".to_string(), self.store.external_host.clone(), self.store.http.listen.port());
+        let url = self.store.external_host.as_ref().map(|x| Url::parse(&format!("https://{}/", x))).and_then(|x| x.ok());
+        let (scheme, host, port) = url
+            .map_or(
+                ("https".to_string(), self.store.external_host.clone(), self.store.http.listen.port()), |
+                x| (x.scheme().to_string(), x.host().map(|x| x.to_string()).or(self.store.external_host.clone()), x.port().unwrap_or(self.store.http.listen.port())));
 
         let (scheme, host, port) = match for_request {
             Some(req) => {
--- a/warpgate-common/src/config/mod.rs
+++ b/warpgate-common/src/config/mod.rs
@@ -141,7 +141,7 @@ pub struct HTTPConfig {
     pub key: String,
 
     #[serde(default)]
-    pub trust_x_forwarded_for: bool,
+    pub trust_x_forwarded_headers: bool,
 }
 
 impl Default for HTTPConfig {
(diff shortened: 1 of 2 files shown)


Eugeny:
Thanks! Would you consider adding support for `X-Forwarded-Proto` in the same PR to cover both bases?

Skyler84:
Add this change to the SSO PR #920? Sure thing!

Skyler84:
- Sure thing, what do you think the `X-Forward-Proto` should affect. I noticed that there should be support for `x-forwarded-...` already but it's all lowercase, but i'm not sure how well this will work as some proxies use `X-Forwarded-...` with upper case words etc

Eugeny:
It's used to construct external URLs for SSO and SSH 2FA. The case doesn't matter because header name matching in both `http` and `poetry` is case insensitive.

Skyler84:
> header name matching in both http and poetry is case insensitive
Ah perfect, I should be able to work on that just fine then. I'll update my bits to use the static header string declarations!

Skyler84:
This works quite nicely now, have tested running behind traefik with multiple hostnames, ports and http/https and the SSO redirect works perfectly. Tested with SSH SSO as well and appears to be working

warpgate-common/src/config/mod.rs
```diff
@@ -295,26 +299,38 @@ impl WarpgateConfig {
...
-                let _ = url.set_port(Some(ext_port));
-            }
-            url
+        // if trust x-forwarded, get x-forwarded-host, then try Host, then fallback on external_host
+        // if trust x-forwarded, get x-forwarded-proto, then try request scheme, then fallback https
+        // if trust x-forwarded, get x-forwarded-port, then try request port, then fallback http listen port
+        let trust_forwarded_headers = self.store.http.trust_x_forwarded_headers;
+        let url = self.store.external_host.as_ref().map(|x| Url::parse(&format!("https://{}/", x))).and_then(|x| x.ok());
+        let (scheme, host, port) = url
+            .map_or(
```
Eugeny, line 308: Struggling to unterstand this line - aren't both branches the same? Did you intend `url` to be something else here?

Skyler84: I realise now that it's probably slightly redundant, as the `Url::parse` will only fail if the warpgate.yaml contains errors. The `map_or` just fallsback in that case, pretty much the sasme as the url parse default values.

Eugeny:
I've reordered things slightly for readability :v:

Thanks!

PR #922, opened 2023-11-29
Make session and cookie age configurable

seysn:
This is a small tweak to make these two fields configurable, which would be really convenient in some situations.


Eugeny:
Looks good to me except please move both to the `http` key 👍

Eugeny:
Thank you!

PR #930, opened 2023-12-26
ci: add arm64 docker image

### LINKED ISSUE #883: Does Warpgate support arm64
Dear Warpgate,

I have tried with code below but docker return "docker: no matching manifest for linux/arm64/v8 in the manifest list entries."
```
docker run --rm -it -v /home/docker/warpgate/data:/data ghcr.io/warp-tech/warpgate:main setup
```

Thank you,
Nguyen

zasdaym:
Resolves #883.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,8 +26,11 @@ jobs:
         with:
           submodules: recursive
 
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

PR #945, opened 2024-02-02
feat(http): add remote_addr to logs

### LINKED ISSUE #944: Client IP in logs to ban
Could it be possible to add the `client_ip` in journald logs so it can be used for example with fail2ban to ban web access attempts? I think should be added in the `log_request_result` function.

NefixEstrada:
With this change, the remote address will be logged. This enables other software, such as fail2ban, to monitor the logs and make actions if required.

Closes #944


Eugeny:
Thank you!

Eugeny:
@all-contributors please add @NefixEstrada for code

PR #957, opened 2024-02-29
Enhance ticket creation api and UI to support ticket expiry

### LINKED ISSUE #924: Feat: add expiry option to tickets
It would be awesome if I could give someone a token with which they can log in for a certain period of time. So only give someone the option to use the ticket for the next 30 minutes. Something that would kind of go along side this would be single use tickets.

kakawait:
Ticket expiry was already supported on core but no way to defined it, neither from UI neither from API

Changed API to accept new optional field `expiry` and update UI form to be able to set it from UI

closes #924

---

UI changes

Screenshot 2024-02-29 at 10 25 12

Screenshot 2024-02-29 at 10 18 50

---

I've not yet implemented ticket _number usage_ because I'm totally new on Rust and Svelte (even TS). I'd like to validate that addition firstly before starting more development.

If ok for you, I could continue PR to add this part


kakawait:
Test failed regarding missing SONAR_TOKEN, I don't think is related at all with code

Eugeny:
Thank you! I've cleaned it up a bit and updated the SDK to make it the request parameter work - added some details about this to the README

Eugeny:
@all-contributors add @kakawait for code

PR #959, opened 2024-03-03
Enhance ticket creation api and UI to support ticket number of usage

kakawait:
Ticket uses left was already supported on core but no way to defined it, neither from UI neither from API

Changed API to accept new optional field  and update UI form to be able to set it from UI

related #924

---

UI changes

Screenshot 2024-03-04 at 00 15 11
Screenshot 2024-03-04 at 00 13 58

### DIFF
--- a/warpgate-admin/src/api/tickets_list.rs
+++ b/warpgate-admin/src/api/tickets_list.rs
@@ -25,7 +25,7 @@ struct CreateTicketRequest {
     username: String,
     target_name: String,
     expiry: Option<DateTime<Utc>>,
-    number_of_usage: Option<i32>
+    number_of_uses: Option<i32>
 }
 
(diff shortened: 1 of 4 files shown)


kakawait:
`numberOfUsage` can be reworded but didn't find better. But also not totally satisfied

Eugeny:
`numberOfUses` would be correct

kakawait:
> `numberOfUses` would be correct

Fixed

In addition, I've added `min=1` for `input type=number` because no sense to be able to add `numberOfUses` lower than 1

Eugeny:
LGTM, thank you!

marco-adeti:
On 0.10.2, the numberOfUses is decreased by 3
If less than 3, can't use ticket. it directly "Used up"

PR #964, opened 2024-03-11
Switch OOB SSO authentication for SSH to use the instructions instead of the name

SheaSmith:
Some SFTP clients (specifically WinSCP) truncate the name parameter for keyboard interactive login (WinSCP specifically puts it in the titlebar):
image

This PR changes the Warpgate authentication flow to use the keyboard interactive parameter for instructions for the primary instructions for SSO web-based auth, instead of the name. This results in the following UI for WinSCP:
image

Filezilla:
image

and OpenSSH:
image

Perhaps some changes to the wording may be in order (e.g. to prevent the duplication of 'Warpgate authentication' in OpenSSH and FileZilla, but I thought I would leave this PR as being as minimal as possible for the meantime.

### DIFF
--- a/warpgate-protocol-ssh/src/server/session.rs
+++ b/warpgate-protocol-ssh/src/server/session.rs
@@ -1360,7 +1360,8 @@ impl ServerSession {
                     login_url.set_fragment(Some(&format!("/login/{auth_state_id}")));
 
                     russh::server::Auth::Partial {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you!

Eugeny:
@all-contributors please add @SheaSmith for code

PR #988, opened 2024-05-16
migrate from moment to date-fns

nosovk:
Last years there are some alternatives to momentjs appeared. Like [date-fns](https://date-fns.org/). Mostly they depend on new [itnl](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Intl), that is build in into all modern browsers.

Unfortunately, I wasn't able to build the project (I've never touched rust before, and the current cargo version refused to accept configs from repo).

That's why I'm not sure that this PR actually works :(

If somebody can check, that it builds correctly now - it would be nice to move to less size costly solution like date-fns.


nosovk:
It seems that sonar has failed because of SONAR_TOKEN env variable missing, it seems that error not related to PR.

Eugeny:
Thank you!

@all-contributors please add @nosovk for code

PR #1021, opened 2024-07-27
PostgreSQL support

### LINKED ISSUE #938: PostgreSQL targets
thanks

Eugeny:
Fixes #938


warpgate-common/src/config/target.rs
```diff
@@ -107,6 +107,24 @@ pub struct TargetMySqlOptions {
     pub tls: Tls,
 }
 
+#[derive(Debug, Deserialize, Serialize, Clone, Object)]
+pub struct TargetPostgresOptions {
+    #[serde(default = "_default_empty_string")]
+    pub host: String,
+
+    #[serde(default = "_default_mysql_port")]
```
NefixEstrada, line 115: This seems to be incorrect!

warpgate-protocol-postgres/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#[async_trait]
+impl ProtocolServer for PostgresProtocolServer {
+    async fn run(self, address: SocketAddr) -> Result<()> {
+        let certificate_and_key = {
+            let config = self.services.config.lock().await;
+            let certificate_path = config
+                .paths_relative_to
+                .join(&config.store.mysql.certificate);
```
NefixEstrada, line 47: This seems wrong

warpgate-protocol-postgres/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+            let key_path = config.paths_relative_to.join(&config.store.mysql.key);
```
NefixEstrada, line 48: This seems wrong

warpgate-protocol-postgres/src/lib.rs
Eugeny: fixed, thanks!

warpgate-protocol-postgres/src/lib.rs
Eugeny: fixed

PR #1049, opened 2024-09-05
Group dependabot version bump prs if they are minor or patch bumps

kekonn:
Based on [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups).

Unfortunately I can't test this, but according to the documentation examples, this should make dependabot group all version bump PRs into a single PR per package manager. Only major version bumps and security issues are still raised as separate PRs.

I also removed the PR limit, since this setting probably makes more sense than simply limiting the dependabot PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,12 +6,24 @@ updates:
   - package-ecosystem: "cargo"
     directory: "/"
     labels: ["type/deps"]
-    open-pull-requests-limit: 25
+    #open-pull-requests-limit: 25
     schedule:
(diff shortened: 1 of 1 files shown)


kekonn:
Incidentally, maybe running these checks should be conditional based on which directories contain changes?

Eugeny:
Let's see if it works :+1:

kekonn:
Well it definitely grouped 16 PRs together for the web version.

Issue #977, opened 2024-03-29
[Enhancement] Show raw hash when configuring OTP


kekonn:
I use Bitwarden to manage my otp so it would be a lot easier for me to copy paste the raw hash used to generate the QR code into Bitwarden, which I can't do now

kekonn:
#1055 has been merged to main

PR #1055, opened 2024-09-10
Validate a TOTP code before saving it

kekonn:
In order to do this, I had to remove the very old and unmaintainted `otplib` and replace it with the well-maintained [otpauth](https://www.npmjs.com/package/otpauth).

I kept the parameters the same, but if there is something only slightly off between the two libraries, this could cause a major headache. I am unsure on how to test this though, or if we can provide some sort of migration for this.

The reason for the upgrade is two-fold:

1. As I said, the original library was unmaintained
2. It's verification mechanism relied on a Buffer class that, as far as I can tell, can't work without some `require` statements that are missing from the library's source files.


kekonn:
This would close #977

PR #1059, opened 2024-09-12
add api usage, and examples for use cli operate targes, roles and users

arstercz:


Eugeny:
Thanks for the PR! 

I would accept a short paragraph in the README that includes the first part about the auth, but not the entire example set, I don't have time to keep them up to date, and the OpenAPI schema already includes all necessary information.

arstercz:
> Thanks for the PR!
> 
> I would accept a short paragraph in the README that includes the first part about the auth, but not the entire example set, I don't have time to keep them up to date, and the OpenAPI schema already includes all necessary information.

hi @Eugeny , some people may need these examples, so I think it might be better to put the `api usage/example set` on the wiki.

Eugeny:
The nightly now includes a full API playground where users can try out requests directly. I'll close this PR.

PR #1180, opened 2024-12-18
fix: create and canonicalize data_path

### LINKED ISSUE #1179: Entering relative path in config leads to UD
The case where you enter a relative path in the config doesn't seem to have been handled properly. It does not migrate the relative path, to an absolute path which it then writes into config, it just writes the relative path, and then on runtime, uses it relatively to where the config file was placed (`/etc`) leading to `./warpdata` becoming `/etc/./warpdata`

Kek5chen:
Fixed #1179


Eugeny:
Thank you!

Eugeny:
@all-contributors add @Kek5chen for code

Issue #177, opened 2022-07-11
Support ssh-agent forwared keys for connection


notnooblord:
Awesome work so far @Eugeny! 
Currently trying to implement warpgate bastion in our org. Been great!

I'm concerned with two things actually:
- to use warpgate everywhere, we would have to drop warpgate public key to each user `.authorized_keys` file, a bit of administrative burden (can be automated)
- warpgate key or server instance gets hacked (somehow) person holding the keys would be able to ssh everywhere bastion can (can be mitigated by ip address limiting, but wont do much if bastion is compromised).

Both can be mitigated if warpgate can support ssh-forwarded keys. So keys for accessing aren't stored on bastion.
Than user can do `ssh -A user:host@wg` and warpgate will use forwarded key to access destination host.

Would such a thing be possible in future?

notnooblord:
I kinda wish in new UI there would be something like this :)

![image](https://user-images.githubusercontent.com/11678665/190533445-42064fe7-4320-4055-9ce4-9b7baf76684c.png)

With forwarded keys not only can those keys be used for auth, but actually this allows bastion host usage as jump-host, it would be really nice feature.

notnooblord:
@Eugeny hi!
Any update on supporting this feature?  

I see that tcp forwarding already been implemented in russh, but I think (could be wrong here) we still need to learn how to talk to ssh agent?

// Maybe possible for me to work on supporting this but don't know where to start :(.

Eugeny:
Not yet - I've added support for agent channels in https://github.com/warp-tech/russh/pull/48, but to couple it with existing agent protocol implementation that russh inherited from thrussh, it needs a bridge between channels and AsyncRead/AsyncWrite traits (also implemented in that PRs).

That latter work was done to implement the SFTP protocol (see https://github.com/warp-tech/russh/issues/89#issuecomment-1342885587), but it should also work for the agent protocol.

If you've got time for it, you can just give it a quick test and see if you're able to e.g. communicate through a forwarded agent channel from a russh server

technologicalMayhem:
Has there been any update on this? I am currently working on a setup where the ability for people to identify themselves using their personal keys instead of the warpgate one would be really useful. I've looked through the (seemingly very out of date) roadmap and could also not find anything similar in the other issues.

This project so far seems really nice and quite promising so I would be sad to have to give this up for a different solution.

notnooblord:
[...]

PR #1249, opened 2025-02-12
Implement Agent Forwarding

samtoxie:
This PR implements Agent forwarding as discussed at https://github.com/warp-tech/warpgate/issues/177#issuecomment-2647526617

Huge thanks to @Eugeny for the pointing me in the right direction


Eugeny:
Perfect 👌

Eugeny:
@all-contributors add @samtoxie for code

aSapien:
Hi guys, how can I enable AgentForwarding auth for a target? I seem to only have Password or Warpgate's Keys.

Warpgate version: 0.15.0

Thanks!

samtoxie:
> Hi guys, how can I enable AgentForwarding auth for a target? I seem to only have Password or Warpgate's Keys.
> 
> Warpgate version: 0.15.0
> 
> Thanks!

Hi aSapien,

Unfortunately that is not implemented in this PR. This only proxies it from client to target, but does not use it for auth to the target.

Im also not quite sure how feasible that will be as (i believe) the necessary channels are only opened after successful auth. Maybe Eugeny can spill his two cents on the feasability. It would be a cool feature though, if I have some spare time soon I might look into it.

Eugeny:
I can't say I'm the biggest fan of the idea as it defeats the centralized credential management, since now you have to provision a specific server's keys to individual users again.

aSapien:
Thanks for the reply! 
To be honest, I came here looking for examples on using [russh](https://github.com/Eugeny/russh) server with Agent Forwarding.
Than, I realized that _this_ project already implements practically 80% of what I am working on. :)

I'm working on creating an SSH proxy server that enables injecting interceptors and rules for detecting and blocking certain traffic signatures. However, in my use case, the proxy (russh instance) should be the only gateway to the upstream. The clients are expected to set up `ForwardAgent` to authenticate and their identity needs to remain intact from the upstream's perspective. 

I will try to figure out how to use the forwarded agent in a russh server, and post my questions in the respective [github repo](https://github.com/Eugeny/russh).

And thank you for a wonderful project, it already provides me with some insights and ideas for my own implementation.

PR #1254, opened 2025-02-13
Implement option to trust unknown audiences

samtoxie:
Currently Warpgate implements openidconnect_rs in a way where the OIDC audience has to be either just the warpgate client id or you have to explicitly trust every additional client id. Some IdP's ([ZITADEL](https://zitadel.com/) in my case) however don't let you limit the audience for a specific client on the IdP side, and in my usecase clients in the IdP project are often added or removed. Having to update the warpgate config every time this happens is undoable for me, as such I would like to explicitly trust any unknown client via config. 

This PR implements that via an added optional setting (`trust_unknown_audiences`) for the OIDC-custom provider.


Eugeny:
Thanks! I'll add it to the wiki

PR #1265, opened 2025-02-20
Fixed #1264 - permission error in kubernetes

### LINKED ISSUE #1264: Permission error with config file in kubernetes
## Description

In Linux systems, file permissions are typically represented using 3 digits, but in some cases they can be represented with 6 digits. For example, when injecting secrets as file in a Kubernetes, permissions are represented using 6 digits.

When secret are mounted as file in kubernetes, 6-digit mode are used (e.g., 100600 instead of 0600).
As a result, attempts to change permissions but fails.

---

**Check file permission in kubernetes**

Tested in Python; returns the same result in Rust

Image

```
>>> import os
>>> print(oct(os.stat('/etc/warpgate.yaml').st_mode))
0o100600
```

**warpgate error message**
Image

**Test Environment**
AWS EKS v1.30.9 / AmazonLinux2023

hashfunc:
Fixed #1264

### DIFF
--- a/warpgate-common/src/helpers/fs.rs
+++ b/warpgate-common/src/helpers/fs.rs
@@ -6,7 +6,7 @@ fn maybe_apply_permissions<P: AsRef<Path>>(
     permissions: std::fs::Permissions,
 ) -> std::io::Result<()> {
     let current = std::fs::metadata(&path)?.permissions();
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks!

Issue #1272, opened 2025-02-26
SSH targets with optional 2fa challenge dont work


samtoxie:
The problem is when a target server has 2fa optionally enabled (for example like so: https://www.digitalocean.com/community/tutorials/how-to-set-up-multi-factor-authentication-for-ssh-on-ubuntu-20-04), as even when a user (such as the one used by warpgate) does not have otp configured the sshd will still force a `keyboard-interactive` auth via the `ChallengeResponseAuthentication` setting required for 2fa. Targets that actually require totp or a different 2nd factor are explicitly part of this report, as those are expected to fail as warpgate cant actually offer the 2nd factor.

In the example below I directly connected to the target using my own key, and not warpgate. This shows the sshd offering `keyboard-interactive` with will not actually prompt anything and instantly succeed:
```
debug1: Offering public key: /home/user/.ssh/id_ed25519 ED25519 SHA256:MYKEY explicit agent
debug1: Server accepts key: /home/user/.ssh/id_ed25519 ED25519 SHA256:MYKEY explicit agent
Authenticated using "publickey" with partial success.
debug1: Authentications that can continue: keyboard-interactive
debug1: Next authentication method: keyboard-interactive
Authenticated to with-optional-otp.my.net ([10.69.224.135]:22) using "keyboard-interactive".
debug1: channel 0: new session [client-session] (inactive timeout: 0)
```

Compared to a target that does not have otp optionally configured at all:
```
debug1: Offering public key: /home/user/.ssh/id_ed25519 ED25519 SHA256:MYKEY agent
debug1: Server accepts key: /home/user/.ssh/id_ed25519 ED25519 SHA256:MYKEY agent
Authenticated to without-optional-otp.my.net ([10.69.218.146]:22) using "publickey".
debug1: channel 0: new session [client-session] (inactive timeout: 0)
```

@Eugeny would it be possible to implement something in the SSH client to just silently accept this (maybe with a 1 or 2 second timeout)? This way warpgate and OTP can be used side by side in different users on a target.

_Originally posted by @samtoxie in https://github.com/warp-tech/warpgate/issues/20#issuecomment-2685457735_

samtoxie:
Ive started working on a fix for this, currently got my quick and dirty attempt working (see https://github.com/samtoxie/warpgate/commit/ccaa385872d3e5ceed9a7ee9557d43ff29960dcd). Now working on structuring it so it can be merged upstream.

Eugeny:
[...]

PR #1273, opened 2025-02-27
Fix SSH Client to respond to keyboard-interactive when target has optional 2FA

samtoxie:
This PR fixes https://github.com/warp-tech/warpgate/issues/1272 for both public key and password targets.

When a keyboard-interactive prompt is offered it will respond with empty responses to given prompts. For targets with enforced 2FA via TOTP for example, this will fail as the code is not correct and the session will terminate. For targets that have 2FA enabled, but don't enforce it this will ensure the connection will be setup as sshd expects those empty responses.

### DIFF
--- a/warpgate-protocol-ssh/src/client/mod.rs
+++ b/warpgate-protocol-ssh/src/client/mod.rs
@@ -565,11 +565,11 @@ impl RemoteClient {
                                 }
                                 let key_str = key.public_key().to_openssh().map_err(russh::Error::from)?;
                                 let mut response  = session
-                                        .authenticate_publickey(
-                                            ssh_options.username.clone(),
-                                            PrivateKeyWithHashAlg::new(key.clone(), best_hash),
-                                        )
-                                        .await?;
+                                    .authenticate_publickey(
+                                        ssh_options.username.clone(),
+                                        PrivateKeyWithHashAlg::new(key.clone(), best_hash),
+                                    )
+                                    .await?;
 
                                 auth_result = self._handle_auth_result(
                                     &mut session,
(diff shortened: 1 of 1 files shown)


samtoxie:
The test failure seems to be because the test ssh server ratelimits after wg attempts kb-interactive. Given the current flow it will always attempt it once per connection if mentioned as possible method and key or password fails.

Ideally you'd pick up on the partial success result, so it only attempts is if pw or key already succeeded individually. However AuthResult does not seem to provide this feedback, any idea for a workaround to still pick that up?

Eugeny:
russh indeed doesn't expose the partial_success flag yet. But I think I've fixed your issue in #1274 - can you please merge or rebase?

Previously Warpgate would try the same RSA key three times with different hash algs, exhausting OpenSSH's auth attempt budget.

samtoxie:
Ah I made a mistake in merging indeed, you caught it quicker than me xD

I wonder if it actually works like this, as at the end we will still worst case attempt n*2 attempts when we also attempt keyboard-interactive for every failed connection. Oh well lets see what the test does.

Eugeny:
It should work in reality except the cases where the auth limit is <=4 or (very artificial) `rsa-sha2-x` signature algos are disabled like in that one test - because right now it ends up with:

* Ed25519 attempt
* K-I attempt
* RSA + (best hash) attempt
* K-I attempt
* ssh-rsa fallback
* (OpenSSH default limit of 5 is here)
* K-I attempt

I'll try to add support for partial_success to russh soon 😅

samtoxie:
> It should work in reality except the cases where the auth limit is <=4 or (very artificial) `rsa-sha2-x` signature algos are disabled like in that one test - because right now it ends up with:
> 
>     * Ed25519 attempt
> 
>     * K-I attempt
> 
>     * RSA + (best hash) attempt
> 
>     * K-I attempt
> 
>     * ssh-rsa fallback
> 
>     * (OpenSSH default limit of 5 is here)
> 
>     * K-I attempt
> 
> 
> I'll try to add support for partial_success to russh soon 😅

Indeed, succeeds now. Thanks for your help! Support for `partial_success` would be awesome indeed, as it would simplify this too. However no rush (no pun intended), I think this is one of the very few cases where it is actually nice to have and for now this workaround is good enough.

Also overall Warpgate really is an awesome project, thanks for all your time and dedication in it :smile:. And its been a nice way so far for me to dig some more into rust, as I usually work with different toolchains.

Eugeny:
Thanks! And thanks for contributing as always

If the stars align I'll have time to try adding partial_success tonight

PR #1299, opened 2025-03-25
Set the config file format

Mice7R:
This sets the file format to be yaml instead of let the config to figure it out by extension. The problem is that it guesses file format by extension and some tools - eg ansible - create tempfiles without extension to validate the config changes before applying them, that makes the validation always fail.

This PR changes how the config file is loaded so that it doesn't guess file type.


Mice7R:
I've changed the code to generate an error instead of defaulting to empty string

Eugeny:
Thank you!

@all-contributors please add @Mice7R for code

Issue #1251, opened 2025-02-12
Delete the known hosts entry when deleting a target


lesslimes:
I was able to notice during the installation of the bastion I had to reinstall a machine which had to carry the same name and same IP and when I wanted to connect me via the warpgate that blocked has all the blow when I was able to notice that the table know host still contained the old key of the preceding machine it would be necessary to remove it at the time of the suppression of the machine on the Web interface and I am on the version v0.12.0

Eugeny:
Hi, did I understand it correctly that you'd like the known host entry to be removed when the target itself is deleted? I'm not sure if you've actually deleted the Warpgate target when reinstalling the machine. If you didn't, you can remove the known host entry from the admin UI without manually editing the database (I'm assuming that's what you did?)

lesslimes:
good evening, to explain what I did I first deleted the virtual machine then I deleted the entry on the web interface of the warpgate I reinstalled the machine I added the new machine on the warpgate with the same name and same IP when connecting it remained blocked on the verification of the know host because it had kept the old in memory or I got it wrong?
(i didn't edit the database manually for the moment)

Eugeny:
Makes sense, thanks for the info! I've updated the title

lesslimes:
thanks for your amazing work !

vallamost:
> good evening, to explain what I did I first deleted the virtual machine then I deleted the entry on the web interface of the warpgate I reinstalled the machine I added the new machine on the warpgate with the same name and same IP when connecting it remained blocked on the verification of the know host because it had kept the old in memory or I got it wrong?
(i didn't edit the database manually for the moment)

This sounds more like a bug than a feature request. The OP deleted the previous host entry using the web UI as intended. Were they supposed to delete the machine a different way to avoid this?

PR #1300, opened 2025-03-27
fix: Remove SSH known hosts entry when deleting SSH targets

Thunderbottom:
When a target is deleted from the admin UI, any associated SSH known hosts entries are now also removed. This prevents orphaned host key entries from remaining in the database after a target is deleted, which could cause confusion if a new target with the same hostname is created later.

closes: #1251

### DIFF
--- a/warpgate-admin/src/api/targets.rs
+++ b/warpgate-admin/src/api/targets.rs
@@ -206,6 +206,18 @@ impl DetailApi {
             .exec(&*db)
             .await?;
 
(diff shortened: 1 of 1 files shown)


Thunderbottom:
I'm fairly new to rust, so do let me know if there's anything that can be improved.

Eugeny:
LGTM 🚀 

@all-contributors add @Thunderbottom for code

PR #1384, opened 2025-06-18
fix(panel): fix user profile link

joseluisgonzalezca:
some strange behaviour can be triggered if you access an HTTP target, go back to homepage and click on the user profile link. By adding `/@wargate` to the URL, the request is always sent to warpgate HTTP panel


Eugeny:
Thank you!

Eugeny:
@all-contributors add @joseluisgonzalezca for code

Issue #923, opened 2023-12-02
Don't verify certificate for secure websocket connections


Emiliaaah:
I have warpgate running in from of a service that has a self signed certificate so it's failing to establish the websocket connection.
When trying to debug why the connection wasn't able to be established this is the error I across in the warpgate logs, which would make me believe warpgate isn't ignoring the certificate (while I do have this specified in the http target config).

```
Error { source: Some(Io(Custom { kind: InvalidData, error: InvalidCertificateData("invalid peer certificate: UnknownIssuer") })) }
```

joseluisgonzalezca:
I think this issue can be closed now

Emiliaaah:
> I think this issue can be closed now

I personally no longer have an install of warpgate running. If you've tested this, and could confirm for me that this is no longer occurring, I'd be more than happy to close the issue.

Emiliaaah:
I'll close the issue for now and reopen if needed

PR #1385, opened 2025-06-18
feat(http): support for insecured websocket connections if `TLS Verify`flag is disabled

joseluisgonzalezca:
Related with issue #923. From the web panel, it is possible to configure a certain target to ignore invalid certificates for HTTP requests. However, this flag was not being applied to websocket connections because they are treated in a different manner.

This fix handles the creation of the TLS connector with the proper configuration in case the flag is enabled.

I'm a noobie with Rust, some maybe some adjustments can be made.

Feel free to give the required feedback.


Eugeny (changes requested):

Eugeny:
Thanks!

Issue #972, opened 2024-03-25
Any credential with SSO and public key credentials available for user asks for password in specific circumstances for SSH


SheaSmith:
If you have the following conditions:

1. A user with an SSH public key and SSO configured.
2. SSH is set to use 'Any credential'
3. You attempt to login to SSH without the specified public key being installed (and therefore would expect to be prompted for a keyboard interactive login)

Then you will be actually asked for a password, rather than the keyboard interactive flow:
![image](https://github.com/warp-tech/warpgate/assets/51303984/5ebb4e55-ddc7-44ed-b405-c1db8e9036c0)

Warpgate config for the user:
![image](https://github.com/warp-tech/warpgate/assets/51303984/e831c780-6638-431a-bc00-9dabb6867d6d)

Happy to provide any relevant logs or config if that helps.

theMackabu:
https://github.com/warp-tech/warpgate/issues/946#issuecomment-2087785920 comment moved here

Eugeny:
I haven't been able to reproduce this but I suspect that your client might have a different preferred auth method order than mine. Anyway, the fix makes sure that password auth won't be offered when the user has no password.

SheaSmith:
Thanks - I'm not seeing the issue with the password any more with the latest update. However, it doesn't seem like I'm prompted with keyboard interactive when I deny public key access (I'm using the 1Password SSH agent, which allows for public key access to be denied) with this config:
![image](https://github.com/user-attachments/assets/3a92e79e-38c7-4341-a053-3b54a3bbebdb)

Here's the logs for `sftp -v` when connecting:
```
sftp -v -P 2222 -o User=shea:<site name> <host>
OpenSSH_for_Windows_8.6p1, LibreSSL 3.4.3
debug1: Authenticator provider $SSH_SK_PROVIDER did not resolve; disabling
debug1: Connecting to <host> [192.168.1.201] port 2222.
debug1: Connection established.
debug1: identity file C:\\Users\\SheaSmith/.ssh/id_rsa type -1
debug1: identity file C:\\Users\\SheaSmith/.ssh/id_rsa-cert type -1
debug1: identity file C:\\Users\\SheaSmith/.ssh/id_dsa type -1
debug1: identity file C:\\Users\\SheaSmith/.ssh/id_dsa-cert type -1
debug1: identity file C:\\Users\\SheaSmith/.ssh/id_ecdsa type -1
[...]

PR #1390, opened 2025-06-23
fix(auth): filter web approval auth method only if there are other authentication methods available

joseluisgonzalezca:
Related to #972.

If you setup a user via SSO and you let "Any credential" as Auth Policy for SSH targets, you will be asked for a password instead of having an SSO link.

This was supposed to be resolved with this commit https://github.com/warp-tech/warpgate/commit/2381f5569670f5b3ea4ce64bd2f97196fadc2fb1. However, a regression was introduced with this other change https://github.com/warp-tech/warpgate/commit/6cb156df1fc83cf68dd84bd36441659f7e332434

With this fix, `WebUserApproval` will always be available for the default policy, unless there are other authentication methods configured for the user.

### DIFF
--- a/warpgate-core/src/config_providers/db.rs
+++ b/warpgate-core/src/config_providers/db.rs
@@ -124,16 +124,18 @@ impl ConfigProvider for DatabaseConfigProvider {
             .copied()
             .collect::<HashSet<_>>();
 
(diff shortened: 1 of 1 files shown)


Eugeny:
LGTM 🚀

Issue #1432, opened 2025-07-18
[ Feature ] Add a ping / healthcheck endpoint in API


hugosxm:
Maybe I did not search well but I can't find a healthcheck endpoint for the API. 
I would be great for kubernetes deployments AND docker image embedded healthchecks

Christian-Little-Tava:
We ended up doing something like this for K8s probes

        livenessProbe:
          path: "/@warpgate/api/info"
          port: 8888
          scheme: HTTPS
        readinessProbe:
          path: "/@warpgate/api/info"
          port: 8888
          scheme: HTTPS

Eugeny:
Added `warpgate healthcheck` set it as the container healthcheck command

Issue #1302, opened 2025-04-01
Add hardening documention to warpgate docs


krishardy:
I intend to add some documentation to warpgate-docs to describe how to harden it: set it up to run as a non-root user, rotate ssh and tls certificates, etc.

Issues aren't turned on for warpgate-docs, so I'm just adding this note here to open a communication channel before I start on the PR.

krishardy:
serviced file: https://github.com/warp-tech/warpgate/issues/1221

PR #1433, opened 2025-07-20
Docker image : add healthcheck, linting and run as regular user

hugosxm:
Hi, 

After reading on that issue :  [#1302](https://github.com/warp-tech/warpgate/issues/1302)

I've made the docker image non root,  I tested creating a user and making ssh connection and it worked.

I also added a healthcheck, but we need a proper healthcheck endpoint for that, see [#1432](https://github.com/warp-tech/warpgate/issues/1432)

And there is some linting / best practices, the only thing i disabled is the hadolint check that wants you to pin version. 
See here : [https://github.com/hadolint/hadolint/wiki/DL3008](hadolint-3008)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -28,6 +28,7 @@ RUN just npm ci \
 
 FROM debian:bullseye-20250520
 LABEL maintainer=heywoodlh
+ARG USER_ID=1000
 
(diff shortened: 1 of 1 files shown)


Eugeny:
Looking good, thank you!

Eugeny:
@all-contributors please add @hugosxm for code

PR #1449, opened 2025-08-03
Fix tcp nodelay for ssh server

tom-90:
In PR #1447, I saw you already added the TCP nodelay options for the different connections. Though I think for the SSH server, because warpgate is setting up it's own TCP listeners, and only calling the `run_stream` function of russh, the `nodelay` option doesn't actually get enabled by the library and that has to be done within warpgate itself.


Eugeny:
Eagle eye 💪 

@all-contributors add @tom-90 for code

PR #1475, opened 2025-09-02
Fallback to email if preferred_username is not available in SSO Provider

SteezyCougar:


nosovk:
Currently, Google SSO fails to create new users, because of missing preferred_username in google IDP.
```yaml
sso_providers:
- name: google
  label: Google login
  auto_create_users: true
```
Configuration like that fails to produce new users.

it would be nice to merge this PR @Eugeny, and publish a new version.

Eugeny:
Sorry for the delay and thank you for the contribution!

Eugeny:
@all-contributors add @SteezyCougar for code

PR #1506, opened 2025-09-30
Correct dependency install command

LarsSven:
The dependency install command in the contribution guide isn't correct. It should be `just npm install` rather than `just npm`.


Eugeny:
Thanks!

PR #1511, opened 2025-09-30
attest builds

Eugeny:


kruton:
@Eugeny this should only be run for versions that will be uploaded (pull requests should be skipped).

Something like:

```
if: github.event_name != 'pull_request'
```

or if you only want to attest certain branches:
```
if: github.event_name != 'pull_request' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/'))
```

Eugeny:
~Why would that be a problem though? An attestation only certifies the build source, not its safety.~ nevermind, I see the token issue

Issue #1288, opened 2025-03-18
Transport endpoint is not connected (os error 107)


wolveix:
Hey! Thanks gain for this great project :)

We're unfortunately semi-regularly encountering an issue where Warpgate starts refusing all SSH traffic. We have to access the server and restart the Docker container for it to come back online. This is all we see in the logs:

```
18.03.2025 22:04:46 ERROR warpgate_protocol_ssh::server: Session failed error=Inactivity timeout
18.03.2025 22:04:46  INFO warpgate_protocol_ssh::server::session: Closed session
18.03.2025 22:04:46  INFO SSH{session=bebda70c-3f8a-4355-820a-f1244ecffc70}: warpgate_protocol_ssh::client: Client session closed
18.03.2025 22:04:46  INFO SSH{session=bebda70c-3f8a-4355-820a-f1244ecffc70}: warpgate_protocol_ssh::client: Closed connection
18.03.2025 22:04:47 ERROR warpgate_protocol_ssh::server: Session failed error=early eof
18.03.2025 22:04:47  INFO warpgate_protocol_ssh::server::session: Closed session
18.03.2025 22:04:47  INFO SSH{session=fffba1a1-536a-4fcc-bdef-3aa667c35e1a}: warpgate_protocol_ssh::client: Client session closed
18.03.2025 22:04:47  INFO SSH{session=fffba1a1-536a-4fcc-bdef-3aa667c35e1a}: warpgate_protocol_ssh::client: Closed connection
18.03.2025 22:04:48 ERROR warpgate::commands::run: Server error error=Transport endpoint is not connected (os error 107)
18.03.2025 23:00:37  INFO warpgate::commands::run: Warpgate version=0.13.1
18.03.2025 23:00:37  INFO warpgate::config: Using config: "/data/warpgate.yaml"
```

I left the first few lines in in case they're useful, but they're just the tail end of the normal barrage of login attempts from attackers. As you can see, there's nothing in the logs from that point until we restart the container. It would be incredibly helpful to both understand what's going wrong, but also if the process could hard exit so that the container would automatically restart.

Let me know if I can provide any further details!

Eugeny:
Could you please run Warpgate with `RUST_LOG=debug` and capture a detailed log around the time of this failure? Thanks

wolveix:
Will do! It may take a few days/weeks for it to occur, as a heads up.

nik-orda:
Hi there!

[...]

PR #1521, opened 2025-10-04
Spawn task immediately when accepting connection

kruton:
To prevent per-connection failures from taking down the entire server, spawn the connection setup into its own task. Failures were seen when teh server accepts a connection but the socket has disconnected before stream.peer_address() was called.

Fixes https://github.com/warp-tech/warpgate/issues/1288


Eugeny:
Thank you!

Issue #1525, opened 2025-10-09
Increasing debug (-d -d -d) does not enable debug logging in russh


kruton:
It seems like increasing the debug level (`-d -d -d`) is intended to [enable debug in `russh`](https://github.com/warp-tech/warpgate/blob/6b988b41a4232355a30698412e758d8ef89c7d51/warpgate/src/logging.rs#L16-L20), but it does not appear to work correctly.

PR #1526, opened 2025-10-10
Bridge log crate to tracing for russh debug output

kruton:
russh uses the log crate for all debug logging, but warpgate uses tracing. Without a bridge, russh debug messages (enabled via -d -d) were silently discarded.

Added tracing-log's LogTracer to forward log crate messages to the tracing subscriber, and propagated initialization errors to the CLI.

Fixes https://github.com/warp-tech/warpgate/issues/1525

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5289,6 +5289,17 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Eugeny:
Thank you!

PR #1527, opened 2025-10-14
Add post-quantum key exchange

kruton:
Updating to the latest russh and adding the post-quantum kex algorithm to the client config.


Eugeny:
Thank you

PR #1539, opened 2025-10-20
Use warpgate.yaml for Container Healthcheck

micha-k:
This pull request implements a health check mechanism for the container using the warpgate.yaml configuration file. Without a specified path to the configuration, the health check currently looks for its settings under /etc/warpgate.yaml. However, if the config path is set as in all examples to /data, the container health status will always remain unhealthy. This change ensures that the health check correctly references the configuration, allowing for accurate health status reporting for the container.


Eugeny:
Thank you!

@all-contributors please add @micha-k for code

PR #1544, opened 2025-10-24
Clean up ssh key generation and loading

LarsSven:
While debugging something on our side with SSH key generation and loading, I noticed there was quite some duplication/complexity in this code. This PR merges duplicate implementations and shortens the implementation of SSH key generation and SSH key loading.


Eugeny:
@all-contributors please add @LarsSven for code

Eugeny:
Thanks!

PR #1549, opened 2025-10-27
Add create-user CLI command

### LINKED ISSUE #1505: [Feature] Add support for setting an admin user without setup
We are trying to deploy warpgate at a large scale fully automated, so we want the full end-to-end setup to be Infrastructure-as-Code without manual steps, as it is part of an automated product that shouldn't require user interaction.

There are two logical ways to do a full Infrastructure-as-Code setup:
1. Using the unattended-setup command from Warpgate
2. Writing a warpgate.yaml config to file from cloud-init, and then running warpgate without doing the setup.

We are running into the issue that unattended-setup doesn't have the flexibility of the yaml config, for example you can't set the TLS certificate path, while the manual config writing without running a setup doesn't set up any admin account I can use to get root access to then configure other accounts.

To do the setup now as IaC, we'd need to first set up a systemd unit to run the setup, and then run another systemd unit to overwrite the config.yaml, which is not a great setup. Ideally we'd like to either be able to set the TLS path in the setup, or be able to set up an initial user independently of the setup process. The second would be preferable as it's hard to keep the unattended-setup fully in-sync with the warpgate.yaml, so we'd prefer to just avoid the limited API of the setup command altogether.

Would it be possible to introduce a new command to the warpgate CLI to create an admin user without having to go through the setup or unattended-setup? Or allow the recovery command to create accounts if they're run against a non-existing account? Alternatively it would also work for us if unattended-setup wouldn't fail if a warpgate.yaml config was already present. Basically any method where we can first write the config, and then establish the admin user after.

LarsSven:
Add a CLI command to create a user in warpgate. This command takes the following arguments:
- Mandatory: username
- Mandatory: password
- Optional: role

This command is then also used by the setup command to simplify the setup command quite a lot (which is quite nice as the setup command is very large).

Our main use case is to create an admin user directly from the CLI, so we would use it like:
> warpgate create-user admin <password> warpgate:admin

I am not very familiar with rust so I may have made some supoptimal decisions around pass-by-value/cloning vs pass-by-reference, please let me know if that part especially looks reasonable.

Resolves #1505

### DIFF
--- a/warpgate/src/commands/create_user.rs
+++ b/warpgate/src/commands/create_user.rs
@@ -1,11 +1,19 @@
+use sea_orm::{ActiveModelTrait, ColumnTrait, EntityTrait, QueryFilter, Set};
 use uuid::Uuid;
-use warpgate_common::{Secret, UserPasswordCredential, UserRequireCredentialsPolicy, WarpgateError};
+use warpgate_common::{
+    Secret, UserPasswordCredential, UserRequireCredentialsPolicy, WarpgateError,
+};
 use warpgate_core::Services;
 use warpgate_db_entities::{PasswordCredential, Role, User, UserRoleAssignment};
-use crate::config::load_config;
-use sea_orm::{ActiveModelTrait, ColumnTrait, EntityTrait, QueryFilter, Set};
 
-pub(crate) async fn command(cli: &crate::Cli, username: &str, password: &Secret<String>, role: &Option<String>) -> anyhow::Result<()> {
+use crate::config::load_config;
+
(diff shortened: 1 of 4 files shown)


LarsSven:
I'm not sure what to do with the failing CI job, it timed out it seems? Is it a flaky test?

LarsSven:
@Eugeny would you have time to review this?

Eugeny:
Sorry for the delay - LGTM, I've changed the CLI interface to take --options and added an option to pass the password via an environment variable

PR #1565, opened 2025-11-05
Add configurable idle timeout for PostgreSQL targets

SteezyCougar:
Currently when web user approval is used with psql (And I assume other ways) if the user is inactive for a few minutes it will force them to login again. This hasn't been a great user experience for some of our users, and so this PR is to fix that issue, along with making the actual timeout configurable allowing administrators to control how long authenticated sessions can remain idle before being terminated. This addresses scenarios where long-lived connections may need to be managed more strictly.

## Changes

- **Configuration**: Added optional `idle_timeout` field to `TargetPostgresOptions` with default value of "10m"
- **TCP Keepalive**: Enabled TCP keepalive on connections to prevent premature disconnections during web auth approval wait
- **Session Monitoring**: Implemented idle timeout checking in the PostgreSQL session loop that tracks activity and closes connections exceeding the configured timeout
- **UI**: Added idle timeout configuration field in the web admin interface for PostgreSQL targets
- **Error Handling**: Invalid timeout values gracefully fall back to default (10 minutes) with warning logs

## Backwards Compatibility
Should be fully backwards compatible!


Eugeny:
Thank you!

PR #1577, opened 2025-11-14
Correct PostreSQL connection string & Allow targets to customize default database name in connection examples

SteezyCougar:
Adds support for targets to specify a custom default database name that will be used in connection string examples shown to users.

### Changes
- Adds `default_database_name` field to target configuration (MySQL and Postgres targets only)
- Updates connection instruction examples to use the custom database name when specified
- Falls back to "database-name" placeholder if not set
- Adds database migration to add the new column

### UI Changes
- Adds input field in target configuration for setting default database name
- Field is only visible for MySQL and Postgres targets
- Connection instructions modal updates reactively when the value changes

### Bug Fix
- Fixes PostgreSQL connection command example to use `-W` flag instead of invalid `--password` flag

### DIFF
--- a/warpgate-web/src/admin/AuthPolicyEditor.svelte
+++ b/warpgate-web/src/admin/AuthPolicyEditor.svelte
@@ -3,6 +3,7 @@ import { Input } from '@sveltestrap/sveltestrap'
 import { CredentialKind, type UserRequireCredentialsPolicy } from './lib/api'
 import type { ExistingCredential } from './CredentialEditor.svelte'
 import InfoBox from 'common/InfoBox.svelte'
+import { SvelteSet } from 'svelte/reactivity'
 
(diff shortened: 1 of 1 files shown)


SteezyCougar:
Now example strings look like:

psql -U 'steezy.cougar@steezycougar.com#STEEZY-DATABASE' --host 'warp.steezycougar.com' --port 55432 -W my-custom-db-name-example

SteezyCougar:
image

SteezyCougar:
I'll revise this PR within the next several days most likely, but for right now it's still WIP

SteezyCougar:
@Eugeny this should be ready for review again! Apologies on taking so long, had a few migration conflicts and then got distracted with some other personal projects

Eugeny:
Thank you! I've moved `defaultDatabaseName` into `TargetOptions` though, where all target type specific options live

PR #1612, opened 2025-12-15
Add SCP example command to SSH targets

LarsSven:
Since SSH usernames when using warpgate often contain `:` and space characters, scp is non-trivial as the default scp command that most people use don't work. It works best with setting the user through `-o`.

Many of our users are confused about this, so it would be very helpful to provide an example command for this in the SSH target. So this PR adds this to warpgate.


LarsSven:
@Eugeny sorry for bothering you, I know you are very busy, but would it be possible to get a review for this PR?

Thanks so much for all your work on this project, it's really helped us a lot at the university.

Eugeny:
Thank you and sorry for the delay

Issue #1618, opened 2025-12-19
Helm - Providing the pre-configured warpgate.yaml via overrides_config cause the local admin user to not be created


MohammedNoureldin:
#### Issue / Limitation:
The current implementation of the Helm chart prevents the `unattended-setup` from being called when the `overrides_config` is defined.

This is basically not wrong, because we do not want to override our predefined configuration. However, this causes the issue that **the default admin is not getting created**, and this will prevent the deployment from being fully-automated.

```
if [ ! -f /data/warpgate.yaml ]; then
  warpgate -c /data/warpgate.yaml  unattended-setup --data-path /data
  ...
fi
```

Another thing I am not happy with when not calling the unattended setup, is that we lose the auto-generated keys and certificates. This is also a potential enhancements, so that we call the unattended-setup anyway to take care of a few things for us, then when done, we replace the generated config with ours, but this is currently out of scope.

#### Possible solution:
We can simply add an `else` to do any steps which are currently missing because of not executing the unattended setup. Though, no missing steps come to my mind except creating the admin user.

I will provide a PR for that.

MohammedNoureldin:
If you, @Eugeny, are OK with this, can we release the 0.19.0 or 0.18.1 soon?

MohammedNoureldin:
Fixed and the corresponding PR merged. This will be part of the next release.

PR #1620, opened 2025-12-19
Create the admin user by default even if unattended-setup not called

MohammedNoureldin:
Related to the issue https://github.com/warp-tech/warpgate/issues/1618


Eugeny:
Thanks! 

@all-contributors please add @MohammedNoureldin for code

PR #1637, opened 2025-12-31
feat(ssh): add configurable client authentication methods

### LINKED ISSUE #1561: No login attempts limiting could be a security risk
While checking our logs, we noticed an IP that was attempting to brute-force a password for our admin user.
Ideally we'd love a way to automatically blacklist ip's after a certain amount of attempts, posting a message to contact an administrator to unblock.

liebermantodd:
Hey folks! First off, thanks for Warpgate - it's been great for managing SSH access to my infrastructure.

## The Problem

I've been running Warpgate exposed on port 2222 and getting absolutely hammered by SSH brute-force attacks. Even with Fail2Ban, the sheer volume of password auth attempts was filling up my SQLite database (got up to 2 million log entries, 600MB, eventually corrupted). 

I wanted to do what I do with regular sshd - just disable password auth entirely so attackers aren't even offered it as an option. But I noticed the auth methods are hardcoded in `run_server`:

```rust
methods: MethodSet::from(&[
    MethodKind::PublicKey,
    MethodKind::Password,
    MethodKind::KeyboardInteractive,
][..])
```

## The Fix

Added three config options to `SshConfig`:

```yaml
ssh:
  client_auth_publickey: true
  client_auth_password: false      # this is what I needed
  client_auth_keyboard_interactive: true
```

All default to `true` so existing configs work unchanged. When you start up you'll see:

```
INFO warpgate_protocol_ssh::server: SSH authentication methods enabled: publickey=true, password=false, keyboard_interactive=true
```

## Testing

I followed your existing test patterns and added `test_ssh_client_auth_config.py` with:

- `TestSshClientAuthPasswordDisabled` - password rejected, pubkey still works
- `TestSshClientAuthPublickeyDisabled` - pubkey rejected, password still works  
- `TestSshClientAuthAllMethodsDisabled` - nothing works when all disabled
- `TestSshClientAuthDefaultConfig` - default config allows everything (backward compat)

Would love if you could run the test suite against this. I tested manually on my own Warpgate instance and it's working well - brute force attackers now get immediately rejected instead of being able to spam password attempts.

Let me know if you'd like any changes to the approach or if I should adjust how the config options are named/structured.

Closes #1561


Eugeny:
Thank you for the contribution! However this needs a few changes:
* I'd like to avoid introducing new config options - instead, these should go into the `Parameters` database model
* They'll need their own checkboxes in `Parameters.svelte`
* Disabling authentication methods should affect the options shown in `AuthPolicyEditor.svelte` correspondingly

liebermantodd:
Hi @eugeny, I've updated the PR to address your feedback:

**Changes made:**
1. **Moved from YAML config to Parameters database** - Settings are now stored in the `Parameters` model and persisted to DB
2. **Added UI toggles in Parameters.svelte** - Admins can enable/disable each SSH auth method (publickey, password, keyboard-interactive)
3. **Updated CredentialEditor.svelte** - When an auth method is disabled globally, the corresponding credential options are hidden from the user policy editor
4. **Added fallback safety** - If all methods are disabled, all are enabled by default (with warning logged)
5. **Added info text in UI** explaining the fallback behavior
6. **Added comprehensive E2E tests** (8 tests covering API + actual SSH auth enforcement)

**Files changed (9):**
- `warpgate-admin/src/api/parameters.rs` - API endpoint
- `warpgate-db-entities/src/Parameters.rs` - DB entity
- `warpgate-db-migrations/` - Migration for new columns
- `warpgate-protocol-ssh/src/server/mod.rs` - SSH auth method enforcement
- `warpgate-web/src/admin/config/Parameters.svelte` - UI toggles
- `warpgate-web/src/admin/CredentialEditor.svelte` - Policy editor filtering
- `warpgate-web/src/admin/lib/openapi-schema.json` - API schema
- `tests/test_ssh_client_auth_config.py` - E2E tests

Tested on a live deployment - password auth is properly blocked when disabled.

liebermantodd:
Also added a small fix: `await` the `api.updateParameters()` call in Parameters.svelte to ensure errors are properly propagated rather than silently ignored.

Eugeny:
Thanks! I've updated the PR to apply options without a server restart as well as added checks to actually reject auth attempts since `config.methods` only affects what the server is advertising.

Note that I've removed the fallback when all options are disabled since it's a bit counterintuitive to the user

Eugeny:
@all-contributors add @liebermantodd for code

PR #1650, opened 2026-01-05
feat(cli): add support for setting config path through env

justinforlenza:
This PR allows for setting the `--config` cli argument via a environment variable.

The rationale behind this is in a docker environment the config is located in `/data/warpgate.yaml` instead of the cli default of `/etc/warpgate.yaml`. Meaning if you want to run any warpgate cli commands using docker exec if you do not provide the `--config` argument you get an error. 

This simplifies docker mamagement by allowing admins to run commands like this:
```
docker exec my-container warpgate client-keys
```
Instead of:
```
docker exec my-container warpgate --config=/data/warpgate.yaml client-keys
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -49,7 +49,7 @@ VOLUME /data
 
 ENV WARPGATE_CONFIG=/data/warpgate.yaml
 
-HEALTHCHECK CMD healthcheck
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks! @all-contributors please add @justinforlenza for code

PR #1656, opened 2026-01-06
Correctly naturally sort all relevant lists in the UI

LarsSven:
https://github.com/warp-tech/warpgate/pull/1645 tried to naturally sort the UI to make the ordering of the targets more intuitive. This however didn't work / wasn't complete for various reasons:
- `compare` is a comparison function factory, not a comparison function, so the way it was called in that PR didn't actually work and as such nothing changed in the sorting order. It still sorted the targets lexicographically
- The PR only naturally sorted targets within a group, but did not naturally sort between groups
- Dropdown boxes for targets also need to be naturally sorted.
- Various other resources like roles, server groups and users within the admin UI were not included in that PR, but could also be naturally sorted.

This PR fixes these issues. I cannot guarantee this PR found all instances that would be good to naturally sort, but I did test all changes against a local warpgate instance that had data that would benefit from both sorting within groups and between groups

### DIFF
--- a/warpgate-web/src/admin/config/targets/Targets.svelte
+++ b/warpgate-web/src/admin/config/targets/Targets.svelte
@@ -51,8 +51,8 @@
                         .thenBy(
                             (a: Target, b: Target) =>
                                 natural(
-                                    (groups.find(g => g.id === a.groupId)?.name ?? "").toLowerCase(),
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you!

PR #1659, opened 2026-01-07
fix: modify GEX parameters in SSH key exchange configuration so that it uses 2048 bits

joseluisgonzalezca:
This PR modifies the default SSH client configuration related to key exchange algorithms.

When using GEX (Group Exchange) algorithms, the underlying SSH library enforces a minimum key size greater than 3000 bits ([reference](https://docs.rs/russh/latest/src/russh/client/mod.rs.html#1656)). This causes the key exchange negotiation to fail on some legacy devices (p.e, MikroTik routers running older versions of RouterOS). These devices only support a maximum of 2048 bits.

With this change, the SSH client configuration is adjusted to allow compatibility with these systems, enabling successful SSH connections while maintaining existing behavior for supported environments.

### DIFF
--- a/warpgate-protocol-ssh/src/client/mod.rs
+++ b/warpgate-protocol-ssh/src/client/mod.rs
@@ -488,12 +488,17 @@ impl RemoteClient {
             Preferred::default()
         };
 
-        let config = russh::client::Config {
+        let mut config = russh::client::Config {
             preferred: algos,
             nodelay: true,
-            gex: russh::client::GexParams::new(2048, 2048, 8192)?,
             ..Default::default()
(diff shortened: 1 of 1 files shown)


Eugeny:
The `russh` GEX group size defaults are based on OpenSSH - but you can make this size reduction only kick in when `allow_insecure_algos` is set

joseluisgonzalezca:
Hi @Eugeny!

I've committed the latest changes with your suggestion. Could you verify that everything is alright?

Thanks!

Eugeny:
Looks good now, thanks!

Issue #1626, opened 2025-12-26
[BUG] Could not access multiple HTTP targets in the same session


joseluisgonzalezca:
Hey!

I'm having some issues while using HTTP targets within the Warpgate project. I’ve noticed that it's not possible for a user within a certain session to access more than one HTTP target at the same time.

### Example scenario

For example, I have configured two HTTP targets:

- WEB-01
- WEB-02

When I initially access WEB-01 target, everything works correctly. However, when I then access the second one (using the same session), any further traffic to WEB-01 is routed to WEB-02 instead.

This behavior makes it impossible for a user to interact with multiple HTTP targets simultaneously. Accessing multiple web applications at the same time is a common usage pattern (e.g. opening several internal tools or dashboards in different tabs). However, I think that Warpgate does not currently handle this scenario correctly.

According to Warpgate's documentation:
> While the target is active, Warpgate will pass-through all HTTP traffic in this session straight to it.

This seems to imply that HTTP routing is session-wide rather than target-specific, which likely explains the behavior. However, this limitation conflicts with normal real-world usage where users frequently need to access several web applications at the same time.

### Actual behavior

- Traffic that was previously routed to one HTTP target is now redirected to the newly accessed one.
- Because of that, only one HTTP target can be effectively used per session.

### Expected (or desired) behaviour 

- A user should be able to interact with multiple HTTP targets simultaneously. This is a common usage pattern while working with web-based tools.

### My temporal workaround

The only workaround I’ve found so far is to open each HTTP target in a separate browser session by using incognito mode. This effectively forces Warpgate to create a new session for each target. While this works for the moment, it’s not a convenient or straightforward solution for users and negatively impacts usability.

Could it be possible to modify the tool to handle this scenario in a proper manner. Is there a better workaround?

Thanks for your time!

alistarle:
Hi, this issue is actually really impacting, some workaround is to use a domain based routing in warpgate, but there is also another bug impacting this feature (the session cookie is not shared making warpgate to not propagate your login)

alistarle:
Looks like this latest commit try to fix this issue again: https://github.com/warp-tech/warpgate/commit/40d4ba34cbd8f87cb0cd962eafd24d56838d7848, maybe give it a try !

Eugeny:
[...]

PR #1553, opened 2025-10-31
Fix cross-domain cookie handling and domain rebinding

SteezyCougar:
This PR fixes cross-domain cookie handling and improves domain rebinding functionality for HTTP targets.

 **1. Cross-Domain Cookie Support**
- Session cookies now work across subdomains of the configured base host (e.g., `warpgate.example.com` and `foo.warpgate.example.com`)
- Cookies are automatically scoped to the base domain with appropriate `Secure` and `SameSite=None` attributes for HTTPS
- Added validation to ensure session cookies only work for the base host and its subdomains, providing security boundaries

**2. Domain Rebinding Improvements**
- When HTTP targets have `external_host` configured, visiting that domain automatically routes to the bound target
- Domain rebinding now takes priority over target selection for authenticated users
- Improved host header detection to work reliably behind proxies

**UI Fixes**
- Fixed embedded UI "Home" button to redirect to warpgate home instead of bound domain home (Otherwise it gets confusing for users if you access a resource on a non-matching bound domain)
- Home button now uses the configured `external_host` to ensure correct navigation

PR #1661, opened 2026-01-08
fix: generate SSO login return URL based on request host

joseluisgonzalezca:
This change modifies how the SSO return URL is generated during the login process so that it takes into account the incoming request and its specific host.

By doing this, we could reach a possible solution to support multiple independent HTTP sessions without interfering with the original behavior that enables cookie sharing across domains (see #1553, #1626 and comment https://github.com/warp-tech/warpgate/issues/1626#issuecomment-3719613249).

Additionally, this approach appears to be the correct one, since during the SSO logout process the return URL construction already takes the request into account.

https://github.com/warp-tech/warpgate/blob/08722341af178f800a26b6ec92923dfd401f4071/warpgate-protocol-http/src/api/sso_provider_list.rs#L303-L317

I’m open to feedback and discussion on whether this is the most appropriate solution, and I’m happy to adapt the implementation if a better approach is suggested.

### DIFF
--- a/warpgate-protocol-http/src/api/sso_provider_detail.rs
+++ b/warpgate-protocol-http/src/api/sso_provider_detail.rs
@@ -61,7 +61,7 @@ impl Api {
             return Ok(StartSsoResponse::NotFound);
         };
         let mut return_url = config
-            .construct_external_url(None, provider_config.return_domain_whitelist.as_deref())?;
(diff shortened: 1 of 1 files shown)


Eugeny:
Looks like I've missed it originally - that's the way it's supposed to work, thanks!

PR #1677, opened 2026-01-20
Improve logging/messaging around SSH target authentication failure

### LINKED ISSUE #1667: Print failure message to terminal when failing to authenticate with target
Currently, when you SSH into a target through warpgate, if the Warpgate public key is not properly loaded into the target, no message is printed and you just get "connection closed". This is both not a great debugging experience when  you run into issues setting up warpgate, but on top of that, for us, we have a lot of students using the targets, and they quite often break their environment by accidentally deleting the public SSH key in their VM.

It would be very helpful for warpgate to actually print something along the lines of "Failed to authenticate with target, make sure the warpgate's credentials are loaded into the target" when the Warpgate -> Target connections fails due to authentication reasons.

LarsSven:
Currently, it is quite opaque when Warpgate fails to authenticate with an SSH target (So this is talking about the Warpgate -> Target SSH connection and not the Client -> Warpgate SSH connection). The following happens:
- A message `Auth failed` is printed to the log
- No message is printed to the user

This can be both hard to understand when you're initially experimenting/setting up Warpgate and make mistakes with authentication, and creates a suboptimal user experience when a user of Warpgate SSHes into the VM and then accidentally deletes the public key of the Warpgate user (which unfortunately happens a bit too often with our users).

This MR makes the following changes:
- Expand log message given in admin logs when this happens
- Add specific message for authentication request failure to print to the user, as it's currently just printed as a generic "Connection Failed"
- Add a small timeout before disconnecting the client as printed error messages are never received by the user, since their session gets disconnected before the message is sent, which means they never receive the message

This is how it looks to the user when warpgate cannot connect to the SSH target now:
Screenshot From 2026-01-20 13-56-00

Resolves #1667

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -5,5 +5,5 @@ rm target/llvm-cov-target/* || true
 cargo llvm-cov clean --workspace
 cargo llvm-cov --no-cfg-coverage-nightly --no-report --workspace --all-features -- --skip agent
 cd tests
-RUST_BACKTRACE=1 ENABLE_COVERAGE=1 poetry run pytest --timeout 60 $@
+RUST_BACKTRACE=1 ENABLE_COVERAGE=1 poetry run pytest --timeout 300 $@
 cargo llvm-cov report --html


LarsSven:
@Eugeny would you perhaps have an idea on how to best solve the async issue in this PR where the two flows race? Since I'm not sure if the 100ms delay is a very clean solution to this problem (though it's definitely an improvement over not receiving the message at all)

Eugeny:
I've spent an hour trying to figure out a clean way to make the session wait for service output events and decided that the 100ms sleep is worth it.

PR #1681, opened 2026-01-22
feat: SFTP access control, enforcement mode, file transfer logging, and user role expiry

### LINKED ISSUE #1679: feat: SFTP access control, file transfer logging, and role expiry
## Background

We use Warpgate internally as our SSH bastion. As our team grew and compliance requirements increased, we needed better control over file transfers, audit trails, and temporary access management.

## Problems

- No way to restrict SFTP file transfers while allowing SSH access
- No audit trail of file transfers for compliance
- Manual tracking of temporary contractor access expiration
- All-or-nothing permissions (no role defaults with per-target exceptions)

### Flexible File Transfer Permissions

**Role Defaults + Target Overrides:**
- Set upload/download permissions at role level (applies to all targets)
- Override for specific targets when needed (inherit/allow/deny)
- Optional constraints: allowed paths, blocked extensions, file size limits

**Example:** Developer role allows uploads everywhere, but production DB overrides to block uploads.

### File Transfer Audit Trail

All SFTP operations logged with:
- Who, what, where, when
- File path, size, SHA256 hash
- Success or denial with reason

### SFTP Enforcement Mode

Instance-wide `strict` / `permissive` toggle:
- **Strict:** Shell, exec, and port forwarding blocked for SFTP-restricted targets
- **Permissive:** SFTP restrictions enforced but shell still works (audit mode)

### Time-Limited Role Assignments

- Assign roles with expiration timestamps
- Expired roles automatically revoked
- Quick presets in UI: 4h, 8h, 12h, 1d, 3d, 7d
- Complete history of role changes

## Implementation

- Admin UI with toggles, dropdowns, and contextual help
- Terraform provider support
- Full backward compatibility (no breaking changes)
- E2E and unit tests covering all scenarios

mrmm:
Closes #1679

## What this does

Adds SFTP file transfer access control, an instance-wide enforcement mode, audit logging, and user role expiry management.

### SFTP Access Control
- **Role defaults** — upload/download toggles, allowed paths, blocked extensions, max file size
- **Target-role overrides** — inherit from role OR explicitly allow/deny per target
- **Hierarchy:** Target override → Role default → System default (allow)

**API:** `GET/PUT /role/{id}/file-transfer`, `GET/PUT /targets/{id}/roles/{role_id}/file-transfer`

### Enforcement Mode (`strict` / `permissive`)
- **Strict:** Shell, exec, port forwarding blocked when SFTP restrictions active
- **Permissive:** SFTP restrictions enforced, shell still works
- SFTP EXTENDED packets handled via allowlist (safe/write/unknown categories)

**API:** `GET/PUT /parameters` (field: `sftp_permission_mode`)

### SCP Removal
Removed the legacy SCP module per maintainer feedback — legacy `scp` commands now pass through as regular exec (blocked in strict mode).

### Audit Logging
All SFTP operations logged: user, target, direction, path, size, SHA256 hash, status, denial reason. Stored in `FileTransferLog` table.

### User Role Expiry
- Optional `expires_at` on role assignments with automatic revocation
- Re-activation via expiry update/removal
- Full history tracking (granted → expiry_changed → revoked → re-enabled)

**API:** `POST/GET/PUT/DELETE /users/{id}/roles/{role_id}/expiry`, `GET /users/{id}/roles/{role_id}/history`

### Admin UI
- Global enforcement mode setting with contextual help
- Role-level file transfer defaults with collapsible advanced restrictions
- Target-level per-role overrides with 3-state controls (inherit/allow/deny)
- User role expiry with countdown display and quick presets (4h–7d)
- Role assignment history timeline

## Design Decisions

- **Inheritance model over explicit booleans** — Target-role permissions default to NULL (inherit from role) rather than explicit true/false. This avoids permission drift when role defaults change and reduces per-target configuration burden.
- **Strict mode as default** — When restrictions exist, strict mode blocks shell/exec/forwarding to prevent SFTP bypass. Permissive mode exists for gradual rollout or audit-only use.
- **SCP removed, not restricted** — SCP's protocol makes reliable interception fragile. Rather than half-working restrictions, SCP was removed entirely. SFTP covers all the same use cases with better protocol support.
- **EXTENDED packet allowlist** — Unknown SFTP extensions are blocked when restrictions are active. Known-safe extensions (statvfs, fsync, limits) always pass through. Write extensions (posix-rename, hardlink) check upload permission.
- **Cumulative max_file_size** — Tracked per Write operation across the stream, not per-packet. Denial happens mid-transfer when the limit is exceeded.
- **Session-scoped permission caching** — Permissions resolved once at session start, not per-packet. Trades real-time revocation for performance.
- **Soft-delete for role revocation** — `revoked_at` timestamp instead of row deletion. Allows re-activation and preserves audit trail integrity.

## Database Migrations

- **`m00031`** — `target_roles`: add `allow_file_upload`, `allow_file_download` (bool, default true), `allowed_paths`, `blocked_extensions` (json), `max_file_size` (bigint)
- **`m00032`** — `parameters`: add `file_transfer_hash_threshold_bytes` (bigint, default 10MB) — controls when SHA256 hashing kicks in for audit logs
- **`m00033`** — `user_roles`: add `granted_at`, `granted_by`, `expires_at`, `revoked_at`, `revoked_by`. Creates `user_role_history` table with indexes on `user_id`, `role_id`, `occurred_at`
- **`m00034`** — `roles`: add `allow_file_upload`, `allow_file_download` (bool, default true), `allowed_paths`, `blocked_extensions` (json), `max_file_size` (bigint) — role-level defaults
- **`m00035`** — `target_roles`: convert `allow_file_upload` / `allow_file_download` from non-nullable bool to nullable bool. Migrates existing `true` → `NULL` (inherit), preserves `false` (explicit deny)
- **`m00036`** — `parameters`: add `sftp_permission_mode` (string, default `"strict"`)

## Breaking Changes

None — defaults match current behavior. Existing targets/roles work as before.

## Related

- Terraform provider: [warp-tech\/terraform-provider-warpgate#18](https://github.com/warp-tech\/terraform-provider-warpgate\/pull\/18)

### Role UI - Upload/Download Toggles
<details>
<summary>Show Screenshot</summary>
Role defaults configuration with toggle switches
</details>

### Target UI - Role Attribution for Upload/Download Override
<details>
<summary>Show Screenshot</summary>
Target overrides with inherit/allow/deny dropdowns
</details>

### User UI - Role Assignment History
<details>
<summary>Show Screenshot</summary>
User role assignment history timeline
</details>

### User UI - Role Expiry Setting
<details>
<summary>Show Screenshot</summary>
Expiry setting popup with presets
</details>

### User UI - Role Expiry Countdown
<details>
<summary>Show Screenshot</summary>
Role expiry countdown display
</details>

### User UI - Make Role Permanent
<details>
<summary>Show Screenshot</summary>
Set role to permanent
</details>

### User UI - Role History
<details>
<summary>Show Screenshot</summary>
Role change history log
</details>


mrmm:
Hello @Eugeny, I am not sure why the test aer failing in the Github Action as I have ran them locally and they passed without an issue. Looks like the issue is 2 different test timeout in CI each time  due to Docker resource exhaustion after ~160 tests (maybe).
I am open to any proposal!

Eugeny:
Thank you for the PR! After a cursory glance, it doesn't seem to actually prevent the user from writing/reading data from the FS as there's still nothing stopping them from doing so from within their shell session?

mrmm:
@Eugeny Thanks fro the feedback, this PR design was mainly around the SCP/SFTP subsystem of SSH, It is possible to intercept the SSH session commands as it is the same-ish mechanism used for recording the SSH session. 
I am planning to look into having a feature that allow the administrator to control/prevent some commands to be ran on the Target server.

This PR also does not cover these use-cases too:
- Allow only the user to upload and/or download file from the server ONLY without SSH session
- Inline file scanning to prevent any malicious file to be uploaded (only file extensions prevention, which was quick-win as a all the metadata were accessible) 

If you think it is mandatory feature to be able to have this in this PR, I will work on it but I will need sometime before being able to deliver this 🙏 

Thanks again fro the review

Eugeny:
I see - unfortunately there's no security gained by this PR as-is unless shell and exec sessions are also forbidden, because there's still (a very easy) way to read/write arbitrary files. So while it might tick the boxes for a specific security audit, it doesn't actually prevent unauthorized access.

It's fundamentally impossible to selectively restrict command execution in shell sessions too, since the SSH server only sees user keystrokes and display output and has no concept of "commands" per se (except exec requests, but again the user can bypass that by simply running commands in the shell).

It's not possible to "recognize" and parse shell input either because that's trivially bypassable by encrypting the commands/inputs, saving it remotely and decrypting it there, so it's never visible on the SSH protocol.

As it is, I'm currently tending towards not accepting this as a feature, however I'm still interested in some parts of this PR, if you could separate them out:
* Time limited role assignments
* SFTP audit logging - please see if it's possible to reuse the russh-sftp crate instead of manually parsing the messages

mrmm:
@Eugeny 
I totally agree with you that this technically does not block file transfer as I can open SSH session and copy/paste the content of a file. 

I will separate the TTL role assignment feature, and look into the usage of rssh-sftp for the SFTP audit logging, for audit log purposes only ! 

While I am at it I added some screenshot to this PR, even if it will not move forward I really appreciate the review 🙏

mrmm:
@Eugeny 
Quick update on the russh-sftp usage (and thanks for pointing that out), it would replace:
- the parser parser.rs
- response.rs
- protocol types in types.rs
---
Follow-up question, would it be acceptable to update this feature to:
- State that this only blocks, SCP, SFTP and RSYNC protocols from uploading but user has SSH this access can bypassed this ?
Which gets me to the next point :
- Is adding the possibility to block interactive SSH sessions which can make the target usable only for File transfer
- This PR supports also the legacy SCP (that does scp -t and scp -f exec) -> if we move forward with the proposal of blocking interactive sessions

Eugeny:
Could you please reformulate the second half of your reply? I couldn't parse it 😓

mrmm:
Sorry @Eugeny for the not clear comment, It was done little bit too lat in the night 😓 

To address this, I suggest adding a Restricted Shell mode to make targets truly "File Transfer Only." Regarding this proposal:

- Interactive Sessions: Should we allow blocking interactive SSH sessions to enforce this file-transfer-only access?
- UI Clarity: For cases where the role blocks file transfer protocols but not the interactive session, I can add a warning in the UI to let admins know the restriction can be bypassed via the shell.
- Legacy SCP: I have already implemented support for legacy SCP (scp -t and -f). Should we keep this to ensure full backward compatibility, or would you prefer I drop it to keep this PR more focused and simplified?

mrmm:
@Eugeny Hello, did you have the time tou check my last comment please ?

Eugeny:
Thanks for getting back to me and sorry for the delay!

> Interactive Sessions: Should we allow blocking interactive SSH sessions to enforce this file-transfer-only access?

Yes - blocking `session` and `exec` channels should be a requirement for using SFTP permissions.

> UI Clarity: For cases where the role blocks file transfer protocols but not the interactive session, I can add a warning in the UI to let admins know the restriction can be bypassed via the shell.

See above - enabling SFTP permissions should not be possible when interactive sessions are allowed.

> Legacy SCP: I have already implemented support for legacy SCP (scp -t and -f). Should we keep this to ensure full backward compatibility, or would you prefer I drop it to keep this PR more focused and simplified?

We can leave these out - no reason to use legacy SCP in 2026.

mrmm:
@Eugeny Thanks a lot for the clarification I will add those requirements then ! 
I will update the PR to match the specs, byt the end of the next week 🙏

mrmm:
@Eugeny Here's an update addressing all your feedback points:

### What changed

**Enforcement mode** (addresses session/exec blocking requirement)
- Added instance-wide `sftp_permission_mode` setting: `strict` (default) / `permissive`
- **Strict mode:** When SFTP restrictions are active on a target, `session`, `exec`, `direct-tcpip`, `direct-streamlocal`, `tcpip-forward`, and `streamlocal-forward` channels are all blocked
- **Permissive mode:** SFTP restrictions enforced, but shell/exec/forwarding remain open (for gradual rollout)
- **File transfer only**: The toggle `file_transfer_only` on role, we can enable this to enforce for that specific role to block the interactive, exec channels and port-forward channels. 

**SCP removed**
- Deleted the entire SCP module per your guidance
- Legacy `scp` commands now pass through as regular exec (blocked in strict mode and if the role with "FIle transfer only" activated)

**SFTP codec replaced**
- Swapped custom SFTP parser with `russh-sftp` codec as you suggested
- Added allowlist-based SFTP EXTENDED packet handling (safe/write/unknown categories)

**Permission inheritance model**
- Role-level defaults: upload/download toggles + advanced restrictions (allowed paths, blocked extensions, max file size)
- Target-role overrides: `null` = inherit from role, explicit `true`/`false` = override
- Hierarchy: Target override → Role default → System default (allow)

**User role expiry**
- Optional `expires_at` on role assignments, automatic revocation, full history tracking

### PR description updated
Rewrote to be more concise with a design decisions section explaining the rationale behind each architectural choice.

### Terraform provider
Companion PR: https://github.com/warp-tech/terraform-provider-warpgate/pull/18``

mrmm:
@Eugeny Please let me know if you think this needs more changed or updated to match the expected behavior.  

Edit: I am planning to rewrite git history to make it easier to grasp as there was some changes and fixes that could have been merge. Let me know if you want me to keep it as is !

mrmm:
Hello @Eugeny sorry for the multiple pings, can you please check this PR 🙏 ?

Eugeny:
Sorry for taking so long - I have a large backlog and can spend very little time on it right now - it might take up to 2 weeks for me to finish a review so please hold tight ✌️

mrmm:
@Eugeny that is no problem at all. I just wanted to make sure that there is no blocker on mu side.
Let me know if there is any issue that I can help with !

Eugeny:
I've merged latest main into the branch and tried reviewing for a while but it's not doable with 10K+ changes in one PR. 

Could you please start by splitting role expiry into its own PR? Then we can go through the rest of the changes one by one

mrmm:
@Eugeny Totally understandable ! I will prepare a total of 2 to 3 PRs depending on your preferable review size ! 
- 1 PR for the Role expiry 
- 1 PR for the File Transfer/SFTP changes 

would that work for you ?

mrmm:
@Eugeny Here is the first PR https://github.com/warp-tech/warpgate/pull/1816  for the role expiry extracting the logic in the same implementation approach 
will prepare SFTP/File transfer related changes later this week !

PR #1712, opened 2026-02-07
Google sso role mapping fix

SteezyCougar:
Summary
 
 TLDR is that auto-role provisioning doesn't work with google oauth out the box since google doesn't natively expose groups. That means we have to use a google service account to do a lookup on the user to see what groups they are part of.

  - Fetch Google Workspace group memberships via Directory API on SSO login (Google doesn't expose groups through
  OIDC)
  - Map groups to roles in config, with support for multiple roles per group: "group@example.com": ["role1", "role2"]
  - Add default_credential_policy on SSO providers to control auth requirements for auto-created users
  - Change missing role during SSO sync from hard error to warning

  Setup

  Requires a GCP service account with domain-wide delegation and admin.directory.group.readonly scope. New optional
  fields on the Google provider: service_account_email, service_account_key, admin_email, role_mappings.

  Backward compatibility

  All new fields are optional. Existing configs work unchanged. RoleMapping uses #[serde(untagged)] so both "group":
  "role" and "group": ["role1", "role2"] work.

  Config example

  sso_providers:
    - name: google
      auto_create_users: true
      default_credential_policy:
        http: ["sso"]
        ssh: ["web"]
        mysql: []
        postgres: ["web"]
      provider:
        type: google
        client_id: "..."
        client_secret: "..."
        service_account_email: "$GOOGLE_SA_EMAIL"
        service_account_key: "$GOOGLE_SA_KEY"
        admin_email: "admin@example.com"
        role_mappings:
          "engineering@example.com": "engineering"
          "admins@example.com": ["warpgate:admin", "engineering"]

### DIFF
--- a/config-schema.json
+++ b/config-schema.json
@@ -398,6 +398,15 @@
                 "null"
               ]
             },
+            "admin_role_mappings": {
+              "type": [
+                "object",
+                "null"
+              ],
+              "additionalProperties": {
+                "$ref": "#/$defs/RoleMapping"
+              }
+            },
             "client_id": {
               "type": "string"
             },
@@ -509,7 +518,7 @@
                 "null"
               ],
               "additionalProperties": {
-                "type": "string"
+                "$ref": "#/$defs/RoleMapping"
               }
             },
             "client_id": {


SteezyCougar:
@Eugeny I think this one is ready! We're running it internally now and it seems to be working!

Eugeny:
Sorry for the delay - I'll be able to test this in a week or so

Christian-Little-Tava:
Any luck @Eugeny ?
Also as always feel free to modify/change as you see fit

SteezyCougar:
Rebased!

PR #1750, opened 2026-03-05
Allow for minimizing the password login menu

LarsSven:
Within our warpgate usecase (and I can imagine there are many users with the exact same usecase), we want all our users to log in through SSO, with the username/password only being used for the admin login by the Warpgate administrator. It doesn't make sense to make the local login so big on the login page for this use case, and it will just cause users to try to enter their SSO credentials in the local login.

This PR adds a new config parameter to `warpgate.yaml`: minimize_password_login

When this value is set to true, the password login is moved to the bottom and hidden behind a collapsible menu. The default value is of course false, in which case nothing changes.

Screenshot From 2026-03-05 12-59-25

Screenshot From 2026-03-05 12-59-35

### DIFF
--- a/warpgate-admin/src/api/parameters.rs
+++ b/warpgate-admin/src/api/parameters.rs
@@ -19,6 +19,7 @@ struct ParameterValues {
     pub ssh_client_auth_publickey: bool,
     pub ssh_client_auth_password: bool,
     pub ssh_client_auth_keyboard_interactive: bool,
+    pub minimize_password_login: bool,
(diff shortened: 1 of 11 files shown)


LarsSven:
I'm not entirely sure what to do with the CI, because the test failure is related to an unrelated kubernetes integration test, and the other job checks that the config is not changed, which it obviously was.

Eugeny:
Thanks, I'll take it from here!

PR #1756, opened 2026-03-06
feat: Add HTTPRoute template to Helm chart

solidassassin:
These changes add the ability to create an `HTTPRoute` resource via `values.yaml`. The Gateway API (to which the `HTTPRoute` resource belongs) is a modern [successor](https://kubernetes.io/docs/concepts/services-networking/gateway/#migrating-from-ingress) to the Ingress API.


Eugeny:
Thank you! @all-contributors add @solidassassin for code

PR #1763, opened 2026-03-07
role-based credential mappings for PostgreSQL targets

modanub:
Allows configuring per-role database credentials on PostgreSQL targets. When a user connects, their Warpgate roles are matched against the target's credential mappings and the first match determines which upstream DB username/password is used. Falls back to the target's default credentials if nothing matches.

Use case: different user groups (admins, developers) connect through the same Warpgate target but get different PostgreSQL privilege levels without needing separate targets.

Changes:
- `PostgresCredentialMapping` struct in target config (role → username/password)
- `get_user_roles()` on `ConfigProvider` trait
- credential selection in the Postgres session's `run_authorized` flow
- admin UI for managing mappings on Postgres targets

### DIFF
--- a/warpgate-web/src/admin/lib/openapi-schema.json
+++ b/warpgate-web/src/admin/lib/openapi-schema.json
@@ -4661,8 +4661,36 @@
           "idle_timeout": {
             "type": "string"
           },
+          "credential_mappings": {
+            "type": "array",
+            "items": {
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you for the contribution! Unfortunately, this is not in the scope of the project. If any connection properties (such as target username) must change in response to the user roles, I prefer there to be multiple targets.

PR #1770, opened 2026-03-10
Polish some Kubernetes UI elements

LarsSven:
This fixes a few "bugs" in the frontend related to Kubernetes:

1. There was no label set for Kubernetes targets in the main menu. It now puts "Kubernetes" as a label on the target.
2. Kubernetes makes use of kubeconfig/certificates, which are large multi-line contents. Warpgate already set a `rows` property, however when floating this doesn't do anything, so you had these very large contents that were within a textbox meant for single-line contents. I now replaced this with style.height, which actually does work when floating.

Screenshot From 2026-03-10 16-36-04

Screenshot From 2026-03-10 16-36-35


Eugeny:
Thanks!

PR #1788, opened 2026-03-16
Bump svelte

### LINKED ISSUE #1768: Frontend stilled darkened and uninteractive when closing menu since 0.21
Since Warpgate 0.21, when for example closing the menu of an SSH target, the background is still darkened and uninteractive.

So normally the menu looks like this:

Image

Let's say you open the SSH menu:

Image

If you now close the menu (close button), it looks like this:

Image

And nothing can be interacted with until you refresh.

LarsSven:
In Warpgate 0.21 a very annoying frontend bug started appearing where when you open a ConnectionDetails target menu in the main warpgate menu, and close it again, the main page would continue to be darkened and uninteractive. Essentially, when you close the menu, the menu would disappear, but the rest of the UI would not get back to an interactive state. This forces your closing a target workflow to be:
1. Close target menu
2. Refresh the page

Which is very annoying

I managed to trace the appearance of this bug back to https://github.com/warp-tech/warpgate/pull/1745, and specifically, the upgrade of Svelte from 5.53.5 to 5.53.7. Reverting this upgrade fixes the bug.

This PR reverts just the svelte upgrade, the other upgrades done in the package-lock.json in that PR have not been reverted. I don't know if you want to revert the full PR or just the part of it that broke the frontend. This PR is just the minimal change to get it to work again.

Resolves #1768

### DIFF
--- a/warpgate-web/package-lock.json
+++ b/warpgate-web/package-lock.json
@@ -43,7 +43,7 @@
         "qrcode": "^1.5.4",
         "rxjs": "^7.8.2",
         "sass": "1.78",
-        "svelte": "^5.53.5",
+        "svelte": "^5.54.0",
         "svelte-check": "^4.4.4",
         "svelte-fa": "^4.0.4",
         "svelte-intersection-observer": "^1.1.1",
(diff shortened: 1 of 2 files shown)


Eugeny:
Thanks - reproduced it with 5.53, but it no longer reproduces in 5.54, so I've just swapped it out for an upgrade instead :+1:

PR #1795, opened 2026-03-18
Automatically generate client certificate when using kubernetes targets

LarsSven:
We've started looking into Kubernetes targets as it is a very interesting feature for us, however currently the UX flow for it is too complicated for our students. Expecting students to create a credential in warpgate, base64 encode them, and then insert them into the kubeconfig is quite an ask, and it's also just a very tedious and manual process, especially if we don't pre-provide them very specific instructions.

What would instead be a much more user-friendly setup is where Warpgate generates the credentials for you, and inserts them itself into the kubeconfig, so that you can just copy-paste the kubeconfig and get to work immediately. The main problem with this approach is that the private key of the certificate is not sent to warpgate when creating a warpgate credential, so we need to locally store it. In this MR, I make use of the browser's IndexedDB. This means that every browser or OS needs to create a new credential, so if someone opens a Kubernetes target on their main PC, a credential gets generated, but then if they do the same on their laptop, another credential gets created. If the laptop credential gets deleted from the PC though, it still all works on the laptop, where it will just correctly create a new credential and not try to reuse its old credential that is no longer valid in warpgate.

I hope you like this approach, for us it would significantly simplify using Warpgate with Kubernetes, and the current flow of manually getting credentials, base64 encoding them and inserting them simply doesn't work for us, we can't deploy Kubernetes targets with that workflow unfortunately.

### DIFF
--- a/warpgate-web/src/common/ConnectionInstructions.svelte
+++ b/warpgate-web/src/common/ConnectionInstructions.svelte
@@ -327,7 +327,6 @@
     save={issueCertificate}
     username={username ?? ''}
     storeInBrowserByDefault={true}
-    closeOnIssue={true}
     onClose={() => { issuingCertificate = false; loadCertificates() }}
 />
 {/if}


Eugeny:
Good idea! It's serviceable, but we need to make the security implications more explicit:

* Private keys should not be silently stored by default when issuing a new certificate from the credential editor. You can add a checkbox saying "Store private key in this browser to automatically generate kubeconfigs later" to the certificate generator modal.
* The connection instructions modal should show a list of certificate credentials as a dropdown, with their corresponding PK status (present / missing in the store) and an option to issue a new one (only if self-service credential management is enabled) 

With those changes, I'd be happy to merge this feature :)

LarsSven:
@Eugeny I agree with those pointers, I've updated it all, trying to optimise the UX as much as possible. I'm pretty happy with how it looks now!

Screenshot From 2026-03-19 12-19-54

Screenshot From 2026-03-19 12-21-03

Screenshot From 2026-03-19 12-21-31

LarsSven:
@Eugeny would it be possible to get this in the 0.22 release? Then we can start using Kubernetes targets for our setups in the next version.

Eugeny:
Thanks! I will get it in 0.22 but it might take me a few more days to do it :v:

LarsSven:
Perfect, thank you!

Eugeny:
I've rearranged the UI a bit, please check that it still makes sense for your use case as it is now

LarsSven:
I just tested it, I really like the UI rework!

LarsSven:
@Eugeny, slightly unrelated to this PR, but would it be possible to get a new release of the Terraform provider sometime soon, so we can actually start testing the Kubernetes targets? :)

Eugeny:
Done :)

Issue #1774, opened 2026-03-11
Postgres app approval: better way to handle multiple connections


SteezyCougar:
We have a requirement that any access tool needs Oauth behind it. Because of that, we are using the postgres connection method where when you put 'any' password, you have to open up the web browser and 'Authorize' the connection.

While this meets our requirements, we are getting a LOT of pushback from the devs who are getting frustrated running certain applications or tools and having to refresh/approve 10+ connections at the same time.

Ideally (I'm not sure exactly how) being able to either approve MULTIPLE connections at once, or more ideally having a grace period where rapid fire subsequent connections can be auto-approved would be ideal.

We have users actually deploying PG-Bouncer locally in order to reduce the amount of approvals they need, which.... Isn't a ideal solution.

Our Devs have made it clear this is a dealbreaker for them, and if we can't find a solution around it then they want our team to find an alternative tool

SteezyCougar:
The more I think about this, I don't think we can safely do this without having something local that is state aware. But that might mean we'd HAVE to have a client of some kind, which I know we want to avoid.

**Problem**
Many orgs require multi-factor access controls on a regular cadence, which means static credentials aren't acceptable. The existing web approval flow handles that requirement, but it breaks down when a client opens many connections simultaneously (we've seen 18+ with some tools). Users end up blindly mass-approving without actually validating each request. Ironically this is less secure than the alternative because an attacker who knows this pattern could slip a connection request in during the flood and it'd likely get approved.

The thing is, the actual requirements usually don't mean EVERY individual connection needs approval. Just that they are short-lived credentials backed by something like SSO login. In a lot of ways, the Tickets workflow handles MOST (But not all) of this exact use-case.

The problem with the existing Ticket system is that only admins can create them. For resources a user already has access to, requiring an admin to issue a ticket daily isn't going to be reasonable.

  **Proposed solution:**
1. User-initiated ticket requests:
users can request a ticket for resources they already have access to (with a configurable option to also request access to resources they DON'T have access to).  Requests should have configurable constraints such as a max duration (e.g. 8h, 12h), max connection count, etc.
2. Configurable auto-approval
ticket requests meeting defined criteria (e.g. user already has the role, duration within limit) can be auto-approved without admin involvement. For requests that don't meet that criteria, it sits somewhere for an admin to approve. (Such as requesting a resource they aren't already granted)
3. Credential expiry flow
[...]

PR #1818, opened 2026-03-23
Self Serve Ticket Requests

SteezyCougar:
This PR Adds a self-service ticket request system where users can request temporary access tickets from
  their profile. The existing ticket system was only usable by admins, and involved an admin seeing the user's credentials (Which isn't always ideal). This PR is meant to be a pretty substantial expansion to the existing ticket system that makes it more useful, and handles a few different use cases.

Changes:

GLOBAL SETTINGS:
1. Adds a new setting called "Allow users to request tickets". When enabled, it also adds the following settings:
    A. Auto-approve when user already has role-based access (I figured this is a easy way to solve https://github.com/warp-tech/warpgate/issues/1774) as then they can just have temporary creds that can be expired whenever. Think similar to how AWS SSO flow works.
    B. Require description on ticket requests
    C. Show all targets in ticket request form (For some resources we can't allow users to have unlimited access. Instead it has to be justified, and approved. Allowing them to see all targets means they can open the requests to these resources, have an admin approve, and then get temporary access. This is probably the most useful part of this PR)
    D. Default max ticket duration (Only applies to requests, but makes it so someone can't request more than your  policies allow)
    E. Max uses per ticket (Similar to above)
image

TARGET SETTINGS
The following settings will now show up in targets (But ONLY if the global setting to allow users to request tickets is enabled)
1. Disable ticket requests for this target (Allows per-resource customization.)
2. Always require admin approval (Can be set for more sensitive resources)
3. Max self-service ticket duration (To set it different than the default)
4. Max users per ticket (To set it different than the default)
image

USER PROFILE
The User profile will now show a "Ticket Requests" section underneath their username:
image

From that ticket requests menu, they can create a request, they can revoke their own active tickets (Not shown but it's below the requests)
image

TICKETS (Admin section)
This has mostly just been expanded to allow approving or denying requests (Deny will ask for a reason), but otherwise should be very similar to what it was before. All existing functionality SHOULD be preserved.


SteezyCougar:
@Eugeny apologies if it's a large PR, but I think it's to a point where it should be ready for you to review. As always feel free to change or modify according to whatever makes the most sense for the app

warpgate-admin/src/api/parameters.rs
```diff
@@ -31,6 +37,12 @@ struct ParameterUpdate {
     pub ssh_client_auth_password: Option<bool>,
     pub ssh_client_auth_keyboard_interactive: Option<bool>,
     pub minimize_password_login: Option<bool>,
+    pub ticket_self_service_enabled: Option<bool>,
+    pub ticket_auto_approve_existing_access: Option<bool>,
+    pub ticket_max_duration_seconds: Option<Option<i64>>,
```
Eugeny, line 42: Does this actually translate into appropriate handling of missing fields when `poem_openapi::Object` gets deserialized? I suspect it just gets flattened into Option<i64> meaning any request from an older client (e.g. terraform) will wipe these fields

warpgate-core/src/config_providers/mod.rs
```diff
@@ -109,6 +110,22 @@ pub async fn authorize_ticket(
                 return Err(WarpgateError::UserNotFound(ticket.username.clone()));
             };
 
+            // Self-service tickets must re-validate role-based access at use time.
+            // This ensures revoking a user's role immediately invalidates their
+            // self-service tickets, unlike admin-created tickets which intentionally
+            // bypass role checks.
+            if ticket.self_service {
```
Eugeny: How does this interact with `ticket_auto_approve_existing_access`? If the user has no access via roles and the ticket is approved, this check still prevents access as far as I see

warpgate-core/src/ticket_requests.rs
```diff
@@ -0,0 +1,487 @@
...
+
+    // Validate description requirement
+    if policy.ticket_require_description && params.description.trim().is_empty() {
+        return Err(CreateTicketRequestError::InvalidInput(
+            "A description is required for ticket requests".into(),
+        ));
+    }
+
+    // Validate description length
+    if params.description.chars().count() > 2000 {
```
Eugeny, line 101: hallucinated?

warpgate-core/src/ticket_requests.rs
```diff
@@ -0,0 +1,487 @@
...
+        .ticket_max_duration_seconds
+        .or(policy.ticket_max_duration_seconds);
+    let effective_duration = match params.duration_seconds {
+        Some(duration) => {
+            if duration <= 0 {
+                return Err(CreateTicketRequestError::InvalidInput(
+                    "Duration must be a positive number".into(),
+                ));
+            }
+            if duration < 60 {
```
Eugeny, line 117: These checks should happen before invalid data can get into the system (at the API boundary)

warpgate-core/src/ticket_requests.rs
```diff
@@ -0,0 +1,487 @@
...
+                    )));
+                }
+            }
+            Some(duration)
+        }
+        None => max_duration, // Default to max when omitted
+    };
+
+    // Validate and enforce uses limits
+    let max_uses = target.ticket_max_uses.or(policy.ticket_max_uses);
```
Eugeny: Is there actually a real world use case for the user self-limiting themselves with the number of uses?

warpgate-core/src/ticket_requests.rs
```diff
@@ -0,0 +1,487 @@
...
+        let request = TicketRequest::ActiveModel {
+            id: Set(request_id),
+            user_id: Set(params.user_id),
+            username: Set(params.username.clone()),
+            target_name: Set(params.target_name.clone()),
+            requested_duration_seconds: Set(effective_duration),
+            requested_uses: Set(effective_uses),
+            description: Set(params.description),
+            status: Set(TicketRequestStatus::Approved),
+            resolved_by_username: Set(Some("system".to_string())),
```
Eugeny: This field should be a nullable `resolved_by_user_id` instead of using a fake username

warpgate-core/src/ticket_requests.rs
```diff
@@ -0,0 +1,487 @@
...
+    let Some(request) = TicketRequest::Entity::find_by_id(request_id)
+        .filter(TicketRequest::Column::Status.eq(TicketRequestStatus::Pending))
+        .one(&*db_conn)
+        .await?
+    else {
+        return Ok(None);
+    };
+
+    // Verify user still exists (use user_id, not username, to avoid confusion if username was reused)
+    let user_exists = warpgate_db_entities::User::Entity::find_by_id(request.user_id)
```
Eugeny, line 267: Needs rebase - the Ticket model is now normalized to use IDs for both user and target

warpgate-core/src/ticket_requests.rs
```diff
@@ -0,0 +1,487 @@
...
+
+    let Some(request) = TicketRequest::Entity::find_by_id(request_id)
+        .filter(TicketRequest::Column::Status.eq(TicketRequestStatus::Pending))
+        .one(&*db_conn)
+        .await?
+    else {
+        return Ok(None);
+    };
+
+    let reason = reason.map(|r| {
```
Eugeny, line 401: same

warpgate-db-entities/src/Ticket.rs
```diff
@@ -11,6 +11,7 @@ pub struct Model {
     #[sea_orm(primary_key, auto_increment = false)]
     pub id: Uuid,
     #[oai(skip)]
+    #[serde(skip_serializing)]
```
Eugeny: Actually the entire `derive(Serialize)` can be removed from this struct

warpgate-db-entities/src/TicketRequest.rs
```diff
@@ -0,0 +1,71 @@
...
+#[oai(rename = "TicketRequestStatus")]
+pub enum TicketRequestStatus {
+    #[sea_orm(string_value = "pending")]
+    Pending,
+    #[sea_orm(string_value = "approved")]
+    Approved,
+    #[sea_orm(string_value = "denied")]
+    Denied,
+    #[sea_orm(string_value = "expired")]
+    Expired,
```
Eugeny, line 18: never used

warpgate-protocol-http/src/api/targets_list.rs
```diff
@@ -55,6 +57,7 @@ impl Api {
         &self,
         ctx: Data<&AuthenticatedRequestContext>,
         search: Query<Option<String>>,
+        for_ticket_request: Query<Option<bool>>,
```
Eugeny: This should be a separate API endpoint with a minimal view into the targets (id + name) - instead of gating every field manually

Eugeny (changes requested):
Thanks for the PR! I've noted a few issues, but generally it looks ok

warpgate-admin/src/api/parameters.rs
Christian-Little-Tava: This is copying the same pattern the upstream (ssh_client_auth_publickey). That said I can do it differently if desired

warpgate-core/src/config_providers/mod.rs
Christian-Little-Tava: I'll push a fix for this!

warpgate-core/src/ticket_requests.rs
Christian-Little-Tava: It seemed like it needed some sort of cap so some user doesn't paste like a ridiculous amount of logs in or something as their justification, but 2,000 is absolutely an arbitrary number. I'm not set on it specifically, but it might be worth having SOME max length so a user doesn't do something dumb

warpgate-core/src/ticket_requests.rs
Christian-Little-Tava: It's put here since it's the shared entry point for both the api and the auto-approve. I could split it up though or add additional validation on the API layer, whatever you think makes the most sense I'm down to do

warpgate-core/src/ticket_requests.rs
Christian-Little-Tava: Honestly, I think it'd be rare but it could be possible. Maybe when pairing with someone and sharing or recording your screen you might want to limit for the sake of confidence? 

That said I can only think of edge-case type reasons here like that, If you think the simplification presents better I'm more than happy to update that

warpgate-protocol-http/src/api/targets_list.rs
Christian-Little-Tava: Ah, good call on that, I'll make that change as well

warpgate-core/src/config_providers/mod.rs
Christian-Little-Tava: Ok, authorize_ticket no longer does a self service role re-check since create_ticket_request has everything we need

warpgate-core/src/ticket_requests.rs
Christian-Little-Tava: Good catch, pushed that change

warpgate-db-entities/src/Ticket.rs
Christian-Little-Tava: Sweet, I removed `Serialize` and `poem_openapi::Object` from Ticket entirely

warpgate-db-entities/src/TicketRequest.rs
Christian-Little-Tava: Removed!

warpgate-protocol-http/src/api/targets_list.rs
Christian-Little-Tava: Done!

SteezyCougar:
@Eugeny just checking in to see if there is anything else

Eugeny:
Finally done - sorry for taking so long!

PR #1819, opened 2026-03-23
 improvements(helm chart):  fix setup job command line argument parsing failure due to trailing backslash and other improvements

SachinMaharana:
- When deploying with 
`helm install warpgate ./helm/warpgate -n warpgate --set image.tag=0.21.1 --set setup.type=job`

 i get the following error:
 ```bash
 error: unexpected argument ' --http-port' found 
 ```

The leading space before --http-port — because \  (backslash + space) is an escaped space character in shell, not a line continuation. 

- also removed hard-coded values from the ingress templates
- added an option to deploy obejcts managed with this helm chart

### DIFF
--- a/helm/warpgate/templates/extra-manifests.yaml
+++ b/helm/warpgate/templates/extra-manifests.yaml
@@ -1,0 +1,8 @@
+{{ range .Values.extraObjects }}
+---
+{{- if typeIs "string" . }}
+{{ tpl . $ }}
+{{ else }}
+{{ tpl (. | toYaml) $ }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


SachinMaharana:
@Eugeny The PR fixes some of the issues i faced while deploying in our environment.

Eugeny:
Thank you!

SachinMaharana:
When you plan to release Warpgate 0.22.0, could you also create a release tag (e.g. chart-vX.Y.Z)? That will trigger the Helm chart publishing. Thanks!

PR #1830, opened 2026-03-31
Send messages to SSH terminal synchronously

LarsSven:
This is a followup of a conversation we had in an earlier PR: https://github.com/warp-tech/warpgate/pull/1677#issuecomment-3801666191

Here we decided to introduce a time::sleep to get around the async issues with closing the SSH connection. This is both ugly and brittle, and only worked for the very specific case that we hacked it into.

I've revisited it and found a better way to do it: There is actually a way to synchronously send messages to the terminal. So this PR does that for the critical places where messages may get lost due to disconnecting the SSH sessions, and then gets rid of the sleep timer.


Eugeny:
Thanks for getting this cleaned up nicely :+1:

Issue #1836, opened 2026-04-03
[FE] Log Admin UI: Proposed Optimization Strategy to Fix Performance Issues


TUSIDENG:
# [FE] Log Admin UI: Proposed Optimization Strategy to Fix Performance Issues

### Context
Currently, the **Log page** in the admin UI becomes unresponsive and causes high CPU/memory usage when the log count grows (even at ~1500+ entries). 

### Observed Behavior
1. **Aggressive Polling**: The frontend issues high-frequency requests to `/api/logs`.
2. **DOM Bloat**: Every new log entry is appended to the UI without a cleanup mechanism. With Svelte 5's fine-grained reactivity, rendering thousands of complex log rows (containing JSON metadata) eventually freezes the browser's main thread.
3. **Memory Leak Risk**: Since older entries aren't destroyed from the frontend state, memory usage increases linearly over time.

#### 1. Frontend Buffer Limit (Circular Buffer)
- **Problem**: Storing an infinite number of log entries in the Svelte store leads to OOM (Out of Memory).
- **Solution**: Implement a hard cap on the frontend log array (e.g., `MAX_LOGS = 500`). When the limit is reached, the oldest entries should be shifted out of the array.

#### 2. DOM Virtualization
- **Problem**: Rendering 1000+ complex components into the DOM is heavy.
- **Solution**: Replace the standard `{#each}` loop with a **Virtual List** component (e.g., `svelte-virtuallists`). This ensures only the ~20 visible rows are rendered in the DOM, regardless of the total log count in memory.

#### 3. Polling & Throttling Improvements
- **Solution**: 
    - Increase the polling interval to reduce CPU/Network overhead.
    - Add a "Clear UI Logs" button to allow users to manually reset the frontend display without truncating the database.

### Environment
- **Warpgate Version**: [v0.21.1]
- **Browser**: [146.0.7680.165]
- **OS**: [Windows 11]

pandeysambhi:
@Eugeny attempted to resolve this, have added a PR. let me know your thoughts

PR #1838, opened 2026-04-04
perf(ui): improve admin log page with virtualization, buffer cap, and calmer polling

pandeysambhi:
Improves admin Log page performance and stability when many entries are loaded, as described in [#1836](https://github.com/warp-tech/warpgate/issues/1836).

**Problem**
Very frequent polling stressed the network and main thread.
Every log line was rendered in the DOM, so the page got heavier as the list grew.
Client-side state could grow without a clear bound, increasing memory pressure.

**Changes**
In-memory cap (MAX_LOGS = 500)
After merges, the list is trimmed: when new logs arrive (prepended), we keep the newest 500; when Load older appends history, we keep the oldest 500 so scrolling into the past still behaves sensibly within the cap.

Virtualized list (@tanstack/svelte-virtual)
Only visible (plus overscan) rows are mounted, with dynamic measurement so variable-height rows stay correct. This targets DOM and render cost rather than only array size.

**Polling**
Interval increased from 1s → 3s to cut request churn while staying reasonably live.

“Clear UI logs”
Clears the on-screen list and reloads the current first page from the API. Does not delete anything in the database.

Request size
PAGE_SIZE for log fetches set to 500 so payloads align with the client buffer.

### DIFF
--- a/warpgate-web/src/admin/log-viewer/LogViewer.svelte
+++ b/warpgate-web/src/admin/log-viewer/LogViewer.svelte
@@ -467,8 +467,10 @@ function parseRichLogEntry(entry: LogEntry): RichLogEntry | null {
                                             {item.text}
                                         </span>
                                         {#each Object.entries(item.values ?? {}) as pair (pair[0])}
+                                        <span class="key-value">
                                             <span class="key">{pair[0]}:</span>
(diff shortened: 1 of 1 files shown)


Eugeny:
Thank you! @all-contributors please add @pandeysambhi for code

PR #1846, opened 2026-04-08
Add support for allowed_ip_range for users

### LINKED ISSUE #1545: Allowed IP range for users
We would be very interested in a feature where we can on a per-user basis configure an IP range that the user is allowed to log in from. This allows us to do things like:
- Admins that can only log in from inside the corporate network
- Machine access to VMs that can only be done from the IPs of the CI runners, even if you manage to extract the machine's SSH key (this is for us very relevant as we have students that we want to give CI access to the VMs, without them themselves being able to log in using the CI's SSH key, that they may be able to extract/leak from their repository).

LarsSven:
This PR adds a new properties to users: allowed_ip_range. If someone tires to log into that user from outside the IP range, the login will be rejected. There are quite a few usecases for this:
1. Companies who want to harden their security by restricting login from their on-site locations.
2. Only allow SSH/kubectl access from the CI (so if someone manages extract the warpgate access token from the CI, they still can't use it)
3. A very niche usecase that we need it for: Running student exams through warpgate, where students should only be logging in from the exam hall.

The main reason to have this over for example firewall-level tightening is that it can be configured per user. So for us for example we don't want students to log in outside the exam hall, but we still want to be able to get in ourselves outside the exam PCs.

I've paid special attention to making sure this works nicely with SSO too. I also tried to get the SSH session to print out a message explaining why the auth was rejected, but I didn't manage to do that, so it just says "password rejected" when the SSH session is rejected due to the IP restriction.

Screenshot From 2026-04-08 16-02-14

Screenshot From 2026-04-08 16-02-44

Closes #1545


LarsSven:
One thing I've realised is that before we merge this, it should be a set of address ranges, and not a single one. That should be changed before merging so we don'r have to break the API

LarsSven:
I've updated the PR to allow for a list of ip ranges, rather than a single IP range. This allows for accepting a set of IPs that cannot be covered by a single range.

Eugeny:
Thanks!

Issue #26, opened 2022-04-17
Add support for SSH certificates


adontz:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-using_openssh_certificate_authentication
https://smallstep.com/blog/use-ssh-certificates/
https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates

kzshantonu:
+1

aderumier:
+1 for this,

it can be used with smallstep step-ca, to generate ssh certificate with oidc .

https://smallstep.com/blog/diy-single-sign-on-for-ssh/

csudderth:
Would love to +1 this. Support for Warp to use a short lived SSH cert as opposed to distributing its SSH keys would be huge.

Eugeny:
Could you all please share the possible advantages of using certificates in case of Warpgate specifically? 

* "just in time"-issued short lived certificates (like step-ca or Teleport) can't be used because that requires a custom/wrapper SSH client 
* certificate auth between Warpgate and the client or between Warpgate and the target would have little functional difference to just using public key auth and won't take advantage of what certificate auth has to offer since it's mostly 1:1 authorization between a client and a server instead of many-to-one

adontz:
> Could you all please share the possible advantages of using certificates in case of Warpgate specifically?


SSH certificates may already be implemented, including a custom/wrapper SSH client, so not supporting them may look like a step back. Your perspective is that warpgate is first and/or primary SSH server people connect to. That's not always the case.

Eugeny:
That's not what I mean. SSH client wrappers that step/teleport use are custom to those services as they need to talk with those services first to issue the temporary certificate.

matthew-salerno:
Both the client to warpgate and warpgate to server connections would benefit from removing the trust-on-first-use pattern the ssh keys require.

Supporting ssh certs from the client<->warpgate would allow for better integration with existing infra. For example, a user could use their custom smallstep wrapper to connect to warpgate.

On the warpgate<->server side, I think it could help with #1297 and similar, as it would allow the dynamic creation of certs without needing a sysadmin to login and change the authorized keys on each server every time.

[...]

PR #1847, opened 2026-04-08
Add Ssh cert auth for targets

Koalab99:
Include implementation of SSH Certificate authentication for targets + tests + user procedure


theredspoon:
Comparing this against #2397 while reviewing that PR (context: #26, #2400).

On the target leg, this is architecturally the same design as #2397: ephemeral per-connection certificate, target trusts via `TrustedUserCAKeys`. The two differ in where the CA key lives, and a few things are worth naming directly rather than leaving implicit:

The CA key here is the existing client auth key, unencrypted PKCS#8 PEM, no passphrase, protected only by filesystem permissions. That means the auth key and the fleet-wide signing capability are the same file, one filesystem read gets both. No `key_id` is set on issued certificates, so unlike #2397, a target's own sshd log doesn't attribute a session to a person. And onboarding is still fully manual TOFU with no rotation or distribution path, the same gap Eugeny flagged in #26 back in June, still true today.

Not raising this to dismiss the work, Eugeny was actively preparing this for merge in April before pausing on it. But given #2397's current momentum and the gaps above, worth asking directly: would you be open to closing this in favor of #2397, with the no-external-dependency case tracked separately as a future pluggable backend, the exact direction you proposed yourself in #26 (https://github.com/warp-tech/warpgate/issues/26#issuecomment-4718278455)?

PR #1851, opened 2026-04-09
fix(ui): resolve config page layout regression caused by flex on main

mrmm:
## Summary

Fixes a layout regression where the Config page nav items appeared vertically centered instead of top-aligned.

## Cause

Commit d14be5d added `display: flex; flex-direction: column` to `<main>` in `App.svelte` for the Log page's virtualized table. This caused `margin: auto` on `.container-max-md` to center content vertically in flex context, breaking Config and all config sub-pages.

## Fix

- **`App.svelte`**: Remove `display: flex; flex-direction: column` from `main`
- **`Log.svelte`**: Wrap content in a `.log-page` flex column container scoped to the Log page only

## Screenshots
- beta version 
Screenshot 2026-04-09 at 15 15 06

- fixed version
image

### DIFF
--- a/warpgate-web/src/admin/App.svelte
+++ b/warpgate-web/src/admin/App.svelte
@@ -101,8 +101,6 @@
 
     main {
         flex: 1 0 0;
-        display: flex;
-        flex-direction: column;
     }
 
     header {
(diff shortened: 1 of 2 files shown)


mrmm:
@Eugeny This is a regression from https://github.com/warp-tech/warpgate/pull/1838 .

Eugeny:
Thank you!

PR #1852, opened 2026-04-09
Add support for disabling the injected menu

LarsSven:
Warpgate injects a session menu for HTTP targets with options to log out and go back to home. For our users this is confusing and they have indicated they prefer to not have this option and just go back using the `@warpgate` URL. Furthermore, for our exam mode we don't want our users to log out so we want to hide this button.


Eugeny:
Thanks!

PR #1853, opened 2026-04-09
Use constant time comparison for admin tokens

LarsSven:
This prevents a very hard but technically possible timing attack


Eugeny:
Thank you!

PR #1868, opened 2026-04-15
fix(http): prioritize ?warpgate-target= query param over host-based domain binding

aav:
### Problem         
When an HTTP target has an external_host matching the request's `Host` header, that domain binding overrides an explicit `?warpgate-target=X` query param. On a domain-bound host, clicking a different  target in the selection page bounces the user back to the domain-bound target - selection appears broken.
                                                                                                                                                                                                          
### Fix
Reorder precedence in get_target_for_request() for SessionAuthorization::User: explicit `?warpgate-target=` query param wins, then host-based external_host match, then session's last target. An explicit request should always win against an inferred one.


Eugeny:
Thanks!

PR #1884, opened 2026-04-27
fix(1883): initialize auth when null on target edit page

kamilkrzeminski:
Targets created via the deprecated `password` field (e.g. by the Terraform provider) have `auth=null`. The edit form crashed with "Cannot read properties of undefined (reading 'kind')" because the template assumed `auth` is always set for Postgres/MySQL targets.

Initialize `auth` to `{ kind: 'Password' }` after loading the target if it is null, so the form can render correctly.

### DIFF
--- a/warpgate-web/src/admin/config/targets/Target.svelte
+++ b/warpgate-web/src/admin/config/targets/Target.svelte
@@ -218,7 +218,15 @@
                 </div>
                 <div class="col">
                     <FormGroup floating label="Authenticate using">
-                        <select class="form-control" bind:value={target.options.auth!.kind}>
+                        <select class="form-control"
+                            value={target.options.auth?.kind ?? 'Password'}
+                            onchange={e => {
+                                const kind = (e.target as HTMLSelectElement).value
+                                target!.options.auth = kind === 'IamRole'
+                                    ? { kind: 'IamRole' }
+                                    : { kind: 'Password', password: target!.options.auth?.password ?? '' }
+                            }}
+                        >
                             <option value="Password">Password</option>
                             {#if $serverInfo?.runningOnEc2}
                                 <option value="IamRole">IAM Role (experimental)</option>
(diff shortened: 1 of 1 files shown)


Eugeny:
I thought this would not be possible - how did you get there? Target options should be normalized to use `auth` [here](https://github.com/warp-tech/warpgate/blob/main/warpgate-admin/src/api/targets.rs#L131) and existing targets should have gotten patched by a [migration](https://github.com/warp-tech/warpgate/blob/main/warpgate-db-migrations/src/m00037_database_target_auth.rs#L51)

kamilkrzeminski:
The Terraform provider builds MySQL/Postgres options using the deprecated flat `password` field rather than the `auth` object. The Go struct has `json:"password,omitempty"`, so when a target is created without a password the field is entirely absent from the JSON payload - no `password`, no `auth` - and the `normalize()` guard in `api_create_target` is a no-op.

But there's a second, more impactful path that affects targets **even when a password is provided**: a perpetual drift cycle.

1. **CREATE** - provider sends `password`, `api_create_target` calls `normalize()` -> `auth` is set correctly.
2. **Read-back** - the API never returns `password` (`skip_serializing` in Rust), so `setTargetOptions` in the provider leaves it out of Terraform state
3. **Next plan** - Terraform sees drift (config has `password`, state doesn't) -> queues an UPDATE
4. **UPDATE** - `api_update_target` does **not** call `normalize()`, so the incoming `password` field is never converted -> `auth = null` is persisted to the DB.

This cycle repeats on every subsequent `terraform apply`, silently clearing `auth` each time.

I've opened a fix on the provider side ([terraform-provider-warpgate#26](https://github.com/warp-tech/terraform-provider-warpgate/pull/26)) that reads the password back from the `auth` object during `setTargetOptions`, eliminating the drift entirely. The frontend guard in this PR remains a useful safety net for any existing affected records in the DB.

Two remaining server-side improvements worth considering:
1. `api_update_target` should call `normalize()` the same way `api_create_target` does
2. `normalize()` could default to `auth = Password { password: "" }` when both fields are `None`, as a guard for any future API clients that omit both.

Eugeny:
I've decided to go with a backend-only approach in #1892 which should uphold the invariant where `auth` is always present. It also runs the migration again to fix up any entires that got reset back to using `password`

Eugeny:
Thank you for investigating this and the PR!

PR #1885, opened 2026-04-27
feat: add jump host ssh option

rjourdan04:
Hi everyone,

When deploying Warpgate into our infrastructure, we faced a problem. We have multiple environments not sharing the same subnet. Each environment has a machine with a public ip that we are using as a "gateway" while the other machines only have private ips.

We didn't want to add complexity our maintenance process so having multiple Warpgate wasn't a solution.
We first tried to do ProxyJump using our client ssh config. It was working as expected but we lost a key feature : Session Recordings as Warpgate was just seeing a TCP flow.

What we wanted was the option to configure server side ProxyJump. This is a key feature for us to deploy Warpgate in our company. 

Before submitting a feature request, we wanted to see if it was possible.
To be honest, we aren't Rust developers so we used AI assistance during development.

We built it and deployed it on a test environment and our needs are 100% satisfied. We are now able to configure ProxyJump for a target, session recordings work perfectly.

We believe this feature could benefit the community.

Let me know if you have any questions or suggestions.

Kind regards,
Romain

### DIFF
--- a/warpgate-admin/src/api/ssh_connection_test.rs
+++ b/warpgate-admin/src/api/ssh_connection_test.rs
@@ -15,9 +15,14 @@ use crate::api::common::require_admin_permission;
 pub struct Api;
 
 #[derive(Object)]
+#[oai(rename_all = "camelCase")]
 struct CheckSshHostKeyRequest {
     host: String,
     port: u16,
+    username: Option<String>,
+    allow_insecure_algos: Option<bool>,
+    auth: Option<SSHTargetAuth>,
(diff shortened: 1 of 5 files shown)


benidk:
Hi @Eugeny ,

I hope you're doing well ! I wanted to gently follow up on this PR.
I understand you're busy. We will be happy to rebase, address feedback, or adjust the approach if needed.
Thanks for all the work you put into Warpgate !

Best,

Eugeny:
Thanks and sorry for the delay - I'm cleaning it up in #2013 now since I can't push to your branch here

@all-contributors please add @rjourdan04 for code

PR #1944, opened 2026-05-17
fix: parse forwarded header lists

immanuwell:
Handles comma-separated `X-Forwarded-*` values when trusted headers are on. Multi-hop proxies do this IRL, AWS ALB appends XFF by default, so yep this is reachable.

Repro:
1. set `http.trust_x_forwarded_headers: true`
2. send a request with `X-Forwarded-Host: public.example, proxy.local`, `X-Forwarded-Proto: https, http`, or `X-Forwarded-For: 203.0.113.10, 10.0.0.2`
3. before this, Warpgate used the whole list. proto could fail parsing and host/logs got messy, not great tbh
4. now it uses the first non-empty value and falls back when the trusted header is empty

Tests: `cargo test -p warpgate-common-http`; `cargo clippy -p warpgate-common-http --all-targets -- -D warnings`; `cargo fmt --check --package warpgate-common-http`

Related: #882, #921, #886

### DIFF
--- a/warpgate-common-http/src/auth.rs
+++ b/warpgate-common-http/src/auth.rs
@@ -7,6 +7,44 @@ use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 use warpgate_common::http_headers::{X_FORWARDED_HOST, X_FORWARDED_PROTO};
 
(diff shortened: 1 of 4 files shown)


Eugeny:
Thank you! @all-contributors please add @immanuwell for code

PR #1946, opened 2026-05-18
Little/max api token duration

SteezyCougar:
We have two use-case that this PR solves

1. We want to ensure no token can last longer than we set our max to be
2. We use a local development tool that we want to make it easy for them to create a token so that the tool can grant them access easier to resources without them having to be involved (Hence why we want to be able to restrict the valid timeframe of said tokens)


warpgate-web/src/gateway/ApiTokenManager.svelte
```diff
@@ -9,22 +10,56 @@
...
     let lastCreatedSecret: string | undefined = $state()
+    let error: string | undefined = $state()
     const now = Date.now()
 
+    const urlParams = new URLSearchParams(get(querystring) ?? '')
+    const autoCreate = urlParams.get('create') === 'true'
+    const paramLabel = urlParams.get('label') ?? ''
+    const paramExpiry = urlParams.get('expiry')
+
+    function parseExpiryParam (val: string | null): number | undefined {
```
Eugeny: please reuse/expand `warpgate-web/src/common/duration.ts`

Issue #1948, opened 2026-05-19
redirect_uri_mismatch when using SSO with target subdomains after #1912


SteezyCougar:
After https://github.com/warp-tech/warpgate/commit/b2e3cf0fc9b0a8a81d17bd1119d275334d42d08b ("read request Host header explicitly", https://github.com/warp-tech/warpgate/pull/1912), the SSO redirect_uri is constructed from the request's Host header via trusted_hostname(). When warpgate is accessed through target-specific subdomains (e.g. grafana.warpgate.example.com), the redirect_uri sent to the OAuth provider becomes https://grafana.warp.example.com/@warpgate/api/sso/return instead of https://warp.example.com/@warpgate/api/sso/return.

OAuth providers (Google, Azure AD, etc.) reject this with redirect_uri_mismatch because only the base warpgate host is registered as a valid redirect URI. Google doesn't support wildcard redirect URIs, and registering every target subdomain individually isn't practical

Previous behavior: construct_external_url lived on WarpgateConfig and, when the request URI didn't yield a usable host (common behind L4 load balancers), fell back to external_host from config. The SSO redirect always used the canonical host.

Current behavior: construct_external_url now reads req.header(HOST) directly, which always exists in HTTP/1.1, so it never falls back to config. The SSO redirect uses whatever subdomain the user happened to access.

Expected behavior: The SSO redirect_uri should always use the canonical external_host from config, regardless of which target subdomain triggered the login flow.

mertcanvural:
This looks like host derivation drift after #1912: the OAuth provider is validating the base Warpgate callback, but the request path is now being rebuilt from the target subdomain host. Easiest way to isolate it is to line up three options side by side — manual diff of emitted redirect_uri vs provider allow-list, a preview proxy like authproxy.dev, or a checker like OAuth Redirect Doctor — and compare the exact return URL built behind the proxy/L4 path. If the emitted URI only differs in host, it strongly points to trusted_hostname/external_host precedence rather than provider config.

Issue #1947, opened 2026-05-19
redirect_uri_mismatch when using SSO with target subdomains after #1912


Christian-Little-Tava:
After b2e3cf0f ("read request Host header explicitly", #1912), the SSO redirect_uri is constructed from the request's Host header via trusted_hostname(). When warpgate is accessed through target-specific subdomains (e.g. grafana.warpgate.example.com), the redirect_uri sent to the OAuth provider becomes https://grafana.warp.example.com/@warpgate/api/sso/return instead of https://warp.example.com/@warpgate/api/sso/return.

OAuth providers (Google, Azure AD, etc.) reject this with redirect_uri_mismatch because only the base warpgate host is registered as a valid redirect URI. Google doesn't support wildcard redirect URIs, and registering every target subdomain individually isn't practical

Previous behavior: construct_external_url lived on WarpgateConfig and, when the request URI didn't yield a usable host (common behind L4 load balancers), fell back to external_host from config. The SSO redirect always used the canonical host.

Current behavior: construct_external_url now reads req.header(HOST) directly, which always exists in HTTP/1.1, so it never falls back to config. The SSO redirect uses whatever subdomain the user happened to access.

Expected behavior: The SSO redirect_uri should always use the canonical external_host from config, regardless of which target subdomain triggered the login flow.

PR #1949, opened 2026-05-19
Fix SSO redirect_uri_mismatch on target subdomains

SteezyCougar:
After upstream #1912 refactored construct_external_url to always read the request Host header, SSO redirects from target subdomains (e.g. grafana.warp.example.com) send the wrong redirect_uri to the OAuth provider. Force the SSO return URL to use external_host from config, which is what gets registered in the OAuth provider.

Ref: [warp-tech/warpgate#1947](https://github.com/warp-tech/warpgate/issues/1948)

### DIFF
--- a/warpgate-protocol-http/src/api/sso_provider_detail.rs
+++ b/warpgate-protocol-http/src/api/sso_provider_detail.rs
@@ -62,7 +62,7 @@ impl Api {
             return Ok(StartSsoResponse::NotFound);
         };
         let mut return_url = construct_external_url(
-            Some(req),
+            None,
             &config,
             provider_config.return_domain_whitelist.as_deref(),
         )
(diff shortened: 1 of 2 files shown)


Eugeny:
Could you please describe your setup in terms of domains? Do your users always get redirected back to the primary domain when they log in?

I'll try to reconcile this with what that change is actually trying to fix: allowing complete SSO flow to happen on the subdomain (with every subdomain having a matching redirect URL registered with SSO)

Eugeny:
Assuming your setup is as you described, I think we'll actually need a config flag to switch the behaviour

SteezyCougar:
Basically

Warpgate runs on `warpgate.my-domain.com`

We use subdomain binding so users can:
1. connect to multiple targets at a time (Users constantly have to be connected to 3 or 4 things at once)
2. Share links that will link to a target and work even if you were pointed at a different target

So that means we have like

Cool-application.warpgate.my-domain.com
Another-one.warpgate.my-domain.com
third-one.warpgate.my-domain.com


That's why we we are having just warpgate.my-domain.com handle the authentication for these sub-domains, otherwise every time we modify or add one we have to go and update it in google settings.

But we have a LOT of these targets, enough that setting every single one as a redirect URL and authorized domain isn't going to be practical

Eugeny:
Adding the option in #1971 - would appreciate if you could give it a quick check - `return_url_domain: external_host` should basically revert it to the previous behaviour

PR #1970, opened 2026-05-24
fix: display security key and browser auth URL in SSH terminal (#1960)

### LINKED ISSUE #1960: Security key not displayed in SSH terminal during browser authorization
### Description
When authenticating via SSH with browser authorization enabled, the security key shown in the browser is not displayed in the SSH terminal. The verification feature is therefore non-functional.

Image

### Expected behaviour
The terminal should display something like:
```
Two-factor authentication
Security key: 6F3D
(user:target@host) password [..]
```
### Actual behaviour
```
Two-factor authentication
(user:target@host) password [..]
```
The security key line is completly absent, thus cannot be verified with the browsers key.

### Tested clients
- OpenSSH (Windows & Mac via Tabby)
- PuTTY (Windows)
- OpenSSH (Ubuntu shell)

None of them display the security key. Verbose output also dosen't show any security key.

### Environment
- [warpgate-v0.23.4-x86_64-linux](https://github.com/warp-tech/warpgate/releases/download/v0.23.4/warpgate-v0.23.4-x86_64-linux) deployed on Ubuntu 26.04 LTS (GNU/Linux 7.0.0-15-generic x86_64) via binary installation.
- Direct connection via IP, no domain was used.

xTamasu:
> **AI-assisted development notice:** This fix was developed with the assistance of Claude (Anthropic). The changes have been manually reviewed by me and intensively tested against a live Warpgate instance. However, independent testing is strongly encouraged before merging.

Fixes #1960.

### Problem

When keyboard-interactive authentication required both TOTP and browser approval simultaneously, the SSH terminal showed no instructions at all. Users had no way to know which URL to open or which security key to verify.

Additionally, the previous implementation stored a `broadcast::Receiver<AuthResult>` in the `WebAuthRequested` state and waited on it with `event.recv().await`. This introduced a race condition: if the browser approval arrived while the user was entering their OTP, the signal could be missed or the await would block indefinitely.

### Changes

**UX improvements**
- The SSH terminal now displays the browser approval URL (when `external_host` is configured) and the spaced security key in the OTP prompt when both TOTP and `WebUserApproval` are required at the same time
- The standalone browser auth flow also shows these details and correctly handles retries and rejections
- URL construction failures now degrade gracefully (banner shown without URL) instead of hard-rejecting the connection

**Race condition fix**
- Replaced `broadcast::Receiver<AuthResult>` in `KeyboardInteractiveState::WebAuthRequested` with a simple `u8` retry counter
- Auth completion is now checked via `try_auth_lazy` on each round instead of waiting on a broadcast channel
- `std::mem::replace` is used to take ownership of the keyboard interactive state before matching, resolving the borrow checker conflict that previously prevented accessing other `self` fields inside match arms

**Retry logic**
- Browser auth attempts are limited to `MAX_RETRIES = 3`; exceeding this rejects the connection and cleans up `auth_state`
- On each retry the terminal shows a "Browser authentication was not confirmed, please try again" message along with a refreshed URL and security key

### Testing

You can test the affected flows directly against a running instance:

```bash
# Browser approval only (WebUserApproval flow)
ssh "browser_auth:server2@178.105.91.152" -p 2222

# OTP flow
ssh "otp_user:server2@178.105.91.152" -p 2222
```

**What to verify:**
- The SSH terminal displays the security key and browser URL in the prompt
- Approving in the browser completes authentication without a race condition
- Rejecting or ignoring the browser request triggers a retry prompt (up to 3 times)
- After 3 failed attempts the connection is rejected cleanly

You will fail at OTP verification on my instance. I don't want to expose my servers to the public, lol.
If someone wants to try on my instance, feel free to drop a message. I'll create a account for you then.


Eugeny:
Thank you for the PR! I've refactored the PR a bit to reduce duplication and to correctly handle both k-i responses. LMK if you notice any issues

@all-contributors add @xTamasu for code

xTamasu:
@Eugeny Thanks! I've set up the version on my servers. Everything looks good so far.

Hope the refactoring wasn't too much of a pain, this was my first time working with rust! 😄

PR #1971, opened 2026-05-25
fixed #1948 - add return_url_domain SSO config option

### LINKED ISSUE #1948: redirect_uri_mismatch when using SSO with target subdomains after #1912

Eugeny:
Adds `SsoProviderConfig::return_url_domain` config option:

* `external_host`: always derive the SSO return URL from the `external_host` config option. Useful if the SSO only has one whitelisted return URL
* `host_header`: prefer to derive the return URL using the request host - useful if each domain is whitelisted individually, and the user can complete the SSO flow without leaving the host

Fixes #1948


SteezyCougar:
This looks like it will fix our issue!

PR #1985, opened 2026-05-29
Audit logging improvements

LarsSven:
We want to make more use of the audit logging, but there were some things missing from the current audit logging for us.

- Authentication logs for people logging in were not included in the audit logs filter, so the audit logs themselves were quite barebone in terms of what they showed (Mostly people starting SSH sessions). This filter now also includes the logs for HTTP authentication.
- Warpgate never logged HTTP authentication failures. These are now also logged
- IPs are added to HTTP authentication
- Improved rich log handling for HTTP authentication logs to pretty up the log entries.
- Prevent duplicate successful-authentication audit entries by emitting the success event only once per accepted auth state.

VOfhu6i_d


warpgate-protocol-http/src/api/auth.rs
```diff
@@ -103,6 +106,29 @@ const PREFERRED_NEED_CRED_ORDER: &[CredentialKind] = &[
     CredentialKind::WebUserApproval,
 ];
 
+fn emit_unknown_authentication_failed_event(
```
Eugeny: Fully duplicated in sso_provider_list.rs

warpgate-protocol-http/src/common.rs
```diff
@@ -223,35 +223,54 @@ pub async fn get_auth_state_for_request(
...
+) -> Result<SessionId, WarpgateError> {
+    let server_handle = server_handle_for_request(req, ctx).await?;
+    Ok(server_handle.lock().await.id())
+}
+
+pub async fn authorize_session(
+    req: &Request,
+    ctx: &UnauthenticatedRequestContext,
+    user_info: AuthStateUserInfo,
+) -> Result<SessionId, WarpgateError> {
```
Eugeny: authorize_session returning SessionId here is a completely random change that has nothing to do with auth

Eugeny (changes requested):

warpgate-protocol-http/src/common.rs
LarsSven: That's fair. This was done because with SSH/MySQL/Postgres, the sessionD was created from the start, but with HTTP it was done lazily so I didn't have access to the sessionId which I needed for the logs.

I agree that this is leaking though, so I'll try to look for a different approach.

LarsSven:
Thank you for the review @Eugeny, I have incorporated the feedback

Eugeny:
Looking good with a few fixes, thanks :+1:

PR #1993, opened 2026-06-02
perf: find targets using defined queries instead of doing it in memory

joseluisgonzalezca:
## Description

This PR improves Warpgate's performance when the backend stores a large number of targets. Previously, when resolving the target associated with an incoming request, Warpgate would query the database for the complete list of targets and then perform the filtering in memory. This approach becomes increasingly inefficient as the number of stored targets grows.

With this change, the filtering is performed directly at the database level, reducing the amount of data retrieved and improving lookup performance. The filter by `hostname` inside the `options` JSON column is a little bit more complex because the query syntax varies based on the underlying backend. I have performed the test for the MySQL backend but I couldn't test it with neither Postgres or SQLite (this should be done).

As a potential future optimization, we could also add a database index on the `name` column of the targets table. In my own Warpgate instance, I have manually created this index and observed additional performance improvements for target lookups.

I'm not proficient with Rust so any improvement or suggestion is welcomed. 

## AI Usage

* [ ] Fully vibe coded
* [x] Human-designed, human-coded (except `hostname` filter queries for Postgres and SQLite)

### DIFF
--- a/warpgate-core/src/config_providers/db.rs
+++ b/warpgate-core/src/config_providers/db.rs
@@ -3,9 +3,10 @@ use std::sync::Arc;
 
 use data_encoding::BASE64;
 use sea_orm::{
-    ActiveModelTrait, ColumnTrait, DatabaseConnection, EntityTrait, ModelTrait, QueryFilter,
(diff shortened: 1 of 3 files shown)


Eugeny:
Thank you! I've improved it a bit in #2019 since your branch is push protected.

PR #2046, opened 2026-06-11
Fix Postgres backend protocol version

PalPat1ne:
## Description

...

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/warpgate-protocol-postgres/src/session.rs
+++ b/warpgate-protocol-postgres/src/session.rs
@@ -417,7 +417,7 @@ impl<S: AsyncRead + AsyncWrite + Send + Unpin> PostgresSession<S> {
         let mut client = match PostgresClient::connect(
             &options,
             ConnectionOptions {
-                protocol_version: ProtocolVersion::PROTOCOL3_2,
+                protocol_version: self.decode_context.protocol_version,
                 parameters: startup.parameters,
             },
         )


Eugeny:
This doesn't fix the underlying issue (the upstream postgres-imitating proxy not complying with Postgres protocol, which includes a mechanism to negotiate version, in particular this is an issue with older pgbouncer versions).

It's intended that Warpgate will try to upgrade the protocol version and translate between 3.0 client and 3.2 server if needed.

I'll provide an UI option to set the preferred version in a separate PR shortly

PalPat1ne:
Thanks for the clarification, that makes sense.

In my case the problematic component is outside of my control, because it is part of the managed PostgreSQL service in Yandex Cloud. So even if the upstream proxy/pgbouncer behavior is technically incorrect, I cannot upgrade or reconfigure it from my side.

An option to set the preferred backend protocol version would solve my use case. I could explicitly force `3.0` for this target and still keep using newer Warpgate versions.

Thanks for looking into this.

PalPat1ne:
Also, as a related UX improvement, it would be very helpful to have a “test connection” button for PostgreSQL targets in the UI.

In this case it was quite hard to understand that the problem was specifically between Warpgate and the PostgreSQL backend, especially because the client-side connection to Warpgate itself looked fine. A connection test from the target settings page could make this much easier to diagnose.

PR #2099, opened 2026-06-23
Warn on unknown/misplaced config keys instead of silently dropping them

### LINKED ISSUE #2098: Unknown / misplaced config keys are silently dropped — `warpgate check` passes green
### Summary

Warpgate's config structs (`WarpgateConfigStore` and friends in `warpgate-common/src/config/mod.rs`) don't use `#[serde(deny_unknown_fields)]`, and `load_config` deserializes with `serde_yaml::from_value` (`warpgate/src/config.rs`). As a result any key Warpgate doesn't recognise — whether a typo or a key nested at the wrong level — is **silently discarded**. The config loads fine, `warpgate check` reports "No problems found", and the setting simply has no effect.

Because there is nothing structurally invalid, `warpgate check` cannot catch it — there's no signal at all that a line was ignored.

### Repro

```yaml
# warpgate.yaml
sso_providers:
  - name: keycloak
    provider:
      type: custom
      client_id: warpgate
      client_secret: secret
      issuer_url: https://id.example/realms/x
      scopes: [openid]
    # WRONG LEVEL: role_mappings is a field of the `custom` provider, so it must
    # be nested under `provider:`. One level up it is an unknown key on the
    # sso_providers entry and is dropped.
    role_mappings:
      admins: warpgate:admin

log:
  retention_days: 90   # not a LogConfig field — the field is `retention`
```

`warpgate check` → `No problems found`. At runtime: SSO logins succeed but **no roles are mapped** (every user lands with nothing), and log retention silently stays at the 7-day default.

### Impact

This is an easy and high-cost footgun, especially for `sso_providers[].provider.role_mappings`, which is one level deeper than people expect. It appears to have bitten several users already:

- #1283 — "Cannot use role_mappings in SSO"
- #1414 — "Unable to use the role_mappings feature with Microsoft Entra and custom sso provider"

(See also #1252 for a related "config silently does the unexpected" surprise around the SSO redirect port.)

### Proposed fix

Keep loading (non-breaking) but **surface** ignored keys. Deserialize through [`serde_ignored`](https://docs.rs/serde_ignored) and `warn!` the full dotted path of every dropped key, so they show up at startup and in `warpgate check`:

```
WARN Ignoring unknown config key `sso_providers.0.role_mappings` — likely a typo or a misplaced key; it has NO effect
WARN Ignoring unknown config key `log.retention_days` — likely a typo or a misplaced key; it has NO effect
```

A stricter follow-up could make `warpgate check` fail on ignored keys (behind a flag, to stay non-breaking for `run`).

I have a small PR implementing the warning approach (with unit tests) ready to open.

ndreno:
Fixes #2098

### Problem

Warpgate's config structs have no `#[serde(deny_unknown_fields)]`, so `load_config`'s `serde_yaml::from_value` silently drops any key it doesn't recognise — a typo, or a key nested at the wrong level. The config loads, `warpgate check` says "No problems found", and the setting has no effect. Two real examples:

- `role_mappings` placed on an `sso_providers` entry instead of nested under `provider:` → SSO logins succeed but no roles are mapped.
- `log.retention_days` instead of `log.retention` → retention silently stays at the default.

`warpgate check` can't catch this because nothing is structurally invalid.

### Change

Deserialize the merged config through [`serde_ignored`](https://docs.rs/serde_ignored) and `warn!` the full dotted path of each ignored key. Loading still succeeds — **non-breaking** — but the dropped keys are now visible at startup and in `warpgate check`:

```
WARN Ignoring unknown config key `sso_providers.0.role_mappings` — likely a typo or a misplaced key; it has NO effect
```

The deserialize+collect step is factored into a small `deserialize_store_collecting_ignored` helper so it can be unit-tested.

### Tests

Added `#[cfg(test)] mod tests` in `warpgate/src/config.rs`:
- known keys produce no ignored paths,
- an unknown top-level key is reported,
- a misplaced nested key (`role_mappings` outside `provider:`) is reported with its full path.

`cargo test -p warpgate`, `cargo fmt --check`, and `cargo clippy` all pass.

### Notes

A natural follow-up (not in this PR, to stay non-breaking) is to let `warpgate check` exit non-zero on ignored keys behind a flag, giving CI a hard gate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6223,6 +6223,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_ignored"
+version = "0.1.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115dffd5f3853e06e746965a20dcbae6ee747ae30b543d91b0e089668bb07798"
+dependencies = [
+ "serde",
+ "serde_core",
+]
+
+[[package]]
 name = "serde_json"
 version = "1.0.150"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -7815,6 +7825,8 @@ dependencies = [
  "schemars 1.2.1",
  "sd-notify",
  "sea-orm",
+ "serde",
+ "serde_ignored",
  "serde_json",
  "serde_yaml",
  "time",
--- a/warpgate/Cargo.toml
+++ b/warpgate/Cargo.toml
@@ -28,6 +28,8 @@ reqwest = { workspace = true, features = ["json"] }
 rustls.workspace = true
 schemars.workspace = true
 sea-orm.workspace = true
+serde.workspace = true
+serde_ignored = "0.1"
 serde_json.workspace = true
 serde_yaml = { version = "0.9", default-features = false }
 time = { version = "0.3", default-features = false }
--- a/warpgate/src/config.rs
+++ b/warpgate/src/config.rs
@@ -29,14 +29,43 @@ pub fn load_config(params: &GlobalParams, secure: bool) -> Result<WarpgateConfig
 
     check_and_migrate_config(&mut store);
 
-    let store: WarpgateConfigStore =
-        serde_yaml::from_value(store).context("Could not load config")?;
+    // Deserialize via serde_ignored so that unknown / misplaced keys are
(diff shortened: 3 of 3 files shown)


Eugeny:
Thanks!

PR #2113, opened 2026-06-28
Add configurable groups_claim for custom OIDC providers

noammeltzer-ax:
### Problem

Some providers (e.g. Okta with an org authorization server) can only emit a `groups` claim and cannot add arbitrarily-named custom claims.
Hence, warpgate can't perform automatic role mapping from `warpgatew_roles` & `warpgate_admin_roles`.

### Proposed solution

Map IdP group memberships from a configurable token claim (e.g. the standard `groups` claim) to Warpgate roles via the existing role_mappings / admin_role_mappings, instead of requiring the IdP to emit dedicated `warpgate_roles` / `warpgate_admin_roles` claims. 

- config: new optional `groups_claim` field on the `custom` SSO provider plus a `groups_claim()` accessor.
- sso: introduce `GroupClaim` / `flatten_group_claim`, accepting a bare string, an array of strings, or SCIM-style objects (RFC 7643) from which both `value` (stable id) and `display` (name) are emitted; results are sorted and de-duplicated. The typed `warpgate_roles` / `warpgate_admin_roles` fields reuse the same parsing.
- request: when `groups_claim` is set, source access and admin groups from that claim (ID token first, then userinfo).
- tests: unit tests for claim flattening (string/array/object/mixed/ dedup/spaces) and e2e tests covering access/admin/combined mapping, spaces, duplicates, unmapped groups, and mapping by id or name.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/config-schema.json
+++ b/config-schema.json
@@ -574,6 +574,13 @@
             "client_secret": {
               "type": "string"
             },
+            "groups_claim": {
+              "description": "OIDC claim to read group memberships from (e.g. \"groups\").\nIts values are mapped to roles via role_mappings / admin_role_mappings.\nWhen unset, the warpgate_roles / warpgate_admin_roles claims are used.",
+              "type": [
+                "string",
+                "null"
+              ]
+            },
             "issuer_url": {
               "type": "string"
             },
--- a/config-schema.json
+++ b/config-schema.json
@@ -568,18 +568,17 @@
                 "$ref": "#/$defs/RoleMapping"
               }
             },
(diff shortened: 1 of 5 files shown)


Eugeny:
I've added two small updates: renamed the config field to `roles_claim` (since `groups` is its value, and not what it is) and added a separate `admin_roles_claim` so that the two can be independent

Eugeny:
Thank you! 

@all-contributors add @noammeltzer-ax for code

PR #2141, opened 2026-07-03
fix(auth): avoid holding the AuthStateStore lock during DB I/O and serialization

### LINKED ISSUE #2134: SSH auth plane stalls under concurrent WebUserApproval logins — global AuthStateStore lock held across DB calls
**Version:** 0.25.5

## Summary

Under a burst of concurrent SSH logins using the `WebUserApproval` flow, new SSH
authentications hang **indefinitely** (no timeout) for as long as a gateway web
UI page stays open. They recover the **moment that page is closed** — or
reloaded, which briefly drops its WebSocket. With **no** web UI page open, the
same burst completes without stalling, and it degrades further as more pages are
open. Already-authenticated **web-ssh (web terminal) sessions are unaffected**
throughout, which localizes the fault to the auth-state machinery rather than the
SSH data plane or the DB itself.

## Core problem (Warpgate-side)

A single global `Mutex<AuthStateStore>` is a serialization point for the whole
SSH auth plane, and it is **held across blocking DB calls**:

1. **`AuthStateStore::create()` holds the global lock across two sequential DB
   round-trips.**
   It is called as `auth_state_store.lock().await.create(...)`, and while
   holding that lock it runs `config_provider…list_users().await` then
   `get_credential_policy().await` (`warpgate-core/src/auth_state_store.rs`).
   Every new SSH auth is therefore serialized behind the global lock for the
   duration of 2 DB queries.

2. **The web UI is a second heavy consumer of the *same* lock.**
   Each open gateway page subscribes to `web_auth_request_signal`;
   `api_get_web_auth_requests_stream` re-acquires the global `auth_state_store`
   lock **per event**, and `get_web_auth_requests` holds it while iterating
   **all** pending states (`all_pending_web_auths_for_user`). Contention grows as
   `O(open_pages × auth_events × pending_auths)` — so the more consoles are open,
   the sooner logins stall.

3. **Reconnect amplifier.** `web_auth_request_signal` is
   `broadcast::channel(100)`; during bursts, laggy subscribers hit
   `RecvError::Lagged` and the loop `while let Ok(id) = rx.recv().await`
   **exits**, closing the WS. Browsers reconnect and each reconnect issues a
   fresh full `get_web_auth_requests` under the lock, feeding the storm.

## Why web-ssh is unaffected

`api_create_web_ssh_session` authorizes against the existing session and starts
the target via `WebSshClientManager`; it never creates an auth state, never uses
`WebUserApproval`, and never locks `auth_state_store`. So it keeps working while
the auth plane is jammed — confirming the bottleneck is the global lock, not
backend health.

## Verification (observed)

- **No web UI page open:** the concurrent-login burst completes with no stall.
- **Web UI page open:** logins hang indefinitely; the backlog drains
  **immediately when the tab is closed** (killing the `web-auth-requests`
  WebSocket), and also on a full page reload (which momentarily drops the same
  WebSocket).

The common factor in every recovery is the disappearance of the page's
`web-auth-requests` WebSocket. This isolates the trigger to that WebSocket
contending on the global `auth_state_store` lock — it is continuous contention,
not a one-shot missed wakeup, and not backend health (web-ssh keeps working).

## The hard point to fix

Stop holding `auth_state_store` across I/O:

- In `create()`, resolve user/policy **before** taking the store lock; hold the
  global lock only for the in-memory map insert.
- In `get_web_auth_requests`, snapshot the `Arc`s and **release the store lock
  before** serializing.
- In the web-auth-requests stream loop, handle `RecvError::Lagged` (continue)
  instead of terminating the WebSocket.

## Reproduction

Keep a gateway page open, then fire ~100 concurrent SSH logins via
`WebUserApproval`, approving each with
`POST /@warpgate/api/auth/state/<id>/approve`. Direct SSH auths hang
indefinitely (no timeout). Closing the tab drains the backlog immediately; a full
reload does too. With no page open, the burst does not stall. Web-ssh stays
healthy throughout.

mba965:
## Description

Fixes #2134.

Under concurrent logins the SSH/HTTP/MySQL/Postgres authentication planes can stall because the global `AuthStateStore` mutex is held across expensive work. This is most visible during `WebUserApproval` flows: while an operator keeps the web-approval page open (which polls `get_web_auth_requests`), new SSH/DB logins can hang until the page is closed, then recover.

Three places held the store lock for too long:

- **`AuthStateStore::create`** performed config-provider database lookups (`list_users`, `get_credential_policy`) and IP-range checks while the caller held the store lock, serializing every new authentication behind those queries.
- **`get_web_auth_requests`** held the store lock while scanning every pending auth state (locking each inner state) and serializing them, so a single open web-auth poll blocked all other auth activity for the duration of the scan.
- **`serialize_auth_state_inner`** held the session-state store lock across a per-session lock, and the web-auth websocket held the store and inner locks across the socket write.

### Changes

- Split `create` into `resolve_user_and_policy` (async; does the database I/O and IP checks **without** the store lock) and a synchronous, infallible `create` that only inserts into the in-memory map. All four callers (SSH, HTTP, MySQL, Postgres) now resolve the user/policy before locking the store.
- Replace `all_pending_web_auths_for_user` with `snapshot_states`, which only clones the `Arc` handles under the store lock; filtering and serialization now happen after the lock is released.
- Serialize auth states and serve the web-auth websocket without holding the store lock across inner locks or socket writes; the websocket loop also tolerates broadcast `Lagged` instead of terminating.

No behavioral change — this purely reduces how long the `AuthStateStore` lock is held.

### Testing

- `cargo check --features mysql,postgres` across the affected crates
- `cargo cranky --all-features` (no new warnings)
- `cargo test -p warpgate-core --all-features` (24 passed)

## AI Usage

* [ ] Fully vibe coded


Eugeny:
Thanks!

PR #2185, opened 2026-07-10
HashiCorp Vault / OpenBao integration for secret management

### LINKED ISSUE #2105: Feature Request: HashiCorp Vault / OpenBao integration for secret management
First of all, thank you to the maintainers and contributors for this great project!

## Summary

Add native integration with HashiCorp Vault and/or OpenBao as a secret backend for Warpgate. Instead of storing target credentials and Warpgate's own ssh keys in the local database / config, Warpgate would resolve them from Vault/OpenBao at runtime.

## Motivation

Warpgate currently stores target secrets (passwords, SSH keys) unencrypted in its own database / config. For teams that already run Vault or OpenBao, this creates a parallel secret store that has to be managed, audited, and rotated separately. A native integration would solve several disadvantages at once.

### 1. Easier secret rotation
Currently, rotating a password or ssh key used by a Warpgate target means updating it in Warpgate in addition to the upstream system. With a Vault/OpenBao backend, the secret can be rotated in one place (or rotated automatically by Vault itself), and Warpgate picks up the new
value on the next connection.

### 2. More secure encrypted storage
Vault/OpenBao provide hardened, audited storage with proper access policies, audit logs, transit encryption, auto-unseal, etc. Delegating secret storage to them is generally stronger than rolling encryption-at-rest inside Warpgate, and it lets security teams apply existing policies and monitoring consistently.

### 3. Sharing one secret across multiple targets
Many targets share credentials in practice (e.g. the same service account used across a fleet of hosts). Today, each Warpgate target carries its own copy of that secret, so rotation means N updates and any drift causes silent failures. With a Vault reference, N targets can point at the same secret path and rotate together atomically.

### 4. Secure storage and rotation of Warpgate's own SSH host/client keys
Warpgate's own private keys (the host keys it presents to clients, and the client keys it uses to connect to upstream SSH targets) are sensitive
long-lived material. Storing them in Vault would make rotation easy instead of a manual, risky operation.

## Proposed approach

- A pluggable "secret backend" abstraction inside Warpgate, with the current DB   storage as the default backend and Vault/OpenBao as an additional one.
- Configuration in `warpgate.yaml` for the backend (address, namespace, auth method).
- Secret fields on targets accept either a literal value (current behaviour) or a reference like `vault://<path>#<field>`.
- Support for the KV v2 engine (static secrets).
- OpenBao should work out of the box since it's API-compatible with Vault.

Simonn21:
## Description

Closes #2105 

## AI Usage

* [ ] Fully vibe coded


Eugeny:
Sorry for taking so long - I'm still overloaded with other tasks and will get to reviewing this ASAP - thanks for contributing!

theredspoon:
Came across this while reviewing #2397, which also adds Vault integration to Warpgate. Different problem, yours relocates static secrets into Vault's KV store, #2397 replaces static SSH credentials with ephemeral certs, but the two collide mechanically in a few places worth knowing about: both add a crate to the workspace, both extend `Services` and `ConnectionError`, both touch `SSHTargetAuth`. If both land as-is, an operator using both features ends up configuring Vault access twice with incompatible key names for the same auth methods.

Worth knowing directly: your `VaultTlsConfig` (CA cert, skip-verify) already solves a real gap in #2397, its Vault client has no TLS option at all, so a private-CA Vault deployment can't work with it today. Your hot-reload path solves another, #2397 builds its Vault client once at startup with no reload. And your configurable AppRole/Kubernetes mount paths solve a third gap, for those two methods specifically. #2397 also has AWS/Azure/GCP auth this PR doesn't have, which would need porting over rather than being solved by consuming your layer as-is.

Given that, the cleaner sequencing looks like: this PR's connection/config layer lands first, and #2397 rebases on top of it rather than shipping a second, parallel Vault client, dropping its own transport/token-cache code, porting its AWS/Azure/GCP auth methods into your `VaultAuthConfig` as new variants, and implementing its certificate-signing call through your existing `vaultrs`-based pattern. That would make this PR strictly more capable too, since it'd pick up cloud workload identity it doesn't currently have.

Flagging this now mainly so the two efforts don't land independently and end up needing a follow-up migration to reconcile.

Simonn21:
@Eugeny I've merged the latest main/upstream changes into my branch and resolved the conflicts introduced by the encryption-at-rest and multi-client-SSH-key features. Ready for another review whenever you have time.

Simonn21:
@Eugeny Just wanted to check in on the status. This feature is actually a blocker for us adopting Warpgate at the moment, so I wanted to ask if there's anything still needed from my side, or if you have a rough idea of when you might get a chance to take a look. No pressure at all, just trying to plan around it. Thanks again for your time!

Eugeny:
Sorry for keeping you waiting - the only reason is my lack of time - I'm working to release admin session approvals right now and unfortunately this causes incoming PRs to pile up.

The PR looks fine to me - also you storing secret refs as prefixed strings already matches nicely with how encrypted at rest creds are stored right now

If this is the only thing holding you off I would suggest going with a local build. Most likely I will be reviewing and merging this in the second-next release cycle, and will likely change the secret backend config to be stored in the DB instead of the config file, plus dropping `//` from the serialized secrets to match encrypted-at-rest password format.

You could go into production and later script automatic DB migration once this PR is merged here

(shameless plug: you can also get a [support package](https://warpgate.null.page/for-business/) and let me do it for you - but full disclosure, it won't get this PR merged faster since the 0.29 scope is already locked in)

Simonn21:
Thank you for your quick and honest response. 

I will consider your approach for a local build.

Eugeny:
I now have time to work on this 🎉  - I'll take over the PR and handle the conflicts

Simonn21:
@Eugeny Thank you for working on this PR. 
I have seen your recent commits and have seen that you removed the functionality that the Warpgate SSH host keys can be stored in vault. Can you explain to me why you did this? 🙂

Eugeny:
Yes - I've thought about it for a long time and decided to leave it out for now. Currently there is no support for the new hostkey preview mechanism (`hostkeys-00@openssh.com` extension), meaning that IRL swapping a host key will lead to no old clients being able to connect due to a key mismatch.

Beyond that, an unreachable Vault will progressively kill the SSH listener of Warpgate nodes one by one as they come up/restart.

Combined with the fact that a compromised warpgate can still simply retrieve secrets any time it wants, there is no upside to retrieving the host keys from the Vault.

I'm open to revisit this if hostkeys-00 gets implemented in the future

PR #2190, opened 2026-07-12
Cache show_session_menu on HTTP requests

LarsSven:
## Description
show_session_menu is implemented quite inefficiently on a very hot path (HTTP requests for HTTP targets). There are 2 problems with the implementation:
1. show_session_menu is queried from the DB on every HTTP request
2. show_session_menu is queried even when it's not needed (non-HTML content, non-200 status codes)

This PR adds an in-memory cache for show_session_menu and short-circuits the logic such that the value is not even looked up if the request is not relevant (non-HTML, non-200). This second part is, of course, barely relevant anymore when you do an in-memory lookup, but it's still a free improvement over unnecessarily reading an atomic bool.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/warpgate-admin/src/api/parameters.rs
+++ b/warpgate-admin/src/api/parameters.rs
@@ -284,6 +284,9 @@ impl Api {
         parameters.analytics_normal = body.analytics_normal.map_or(NotSet, Set);
 
         Parameters::Entity::update(parameters).exec(&*db).await?;
+        if let Some(show_session_menu) = body.show_session_menu {
+            services.set_show_session_menu(show_session_menu);
+        }
         drop(db);
 
         services
--- a/warpgate-core/src/services.rs
+++ b/warpgate-core/src/services.rs
@@ -1,10 +1,10 @@
 use std::net::IpAddr;
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 
 use anyhow::Result;
-use sea_orm::DatabaseConnection;
+use sea_orm::{ConnectionTrait, DatabaseBackend, DatabaseConnection};
 use tokio::sync::Mutex;
 use tracing::warn;
 use uuid::Uuid;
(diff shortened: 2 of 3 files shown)


LarsSven:
In my own testing this had a 30.000x speedup specifically for show_session_menu querying (which makes sense, you're comparing a disk read against a memory lookup), though this was on a shared Windows/WSL disk, which is notoriously slow. So the speedup should be lower on a normal linux machine.

LarsSven:
One thing I'm still concerned about with this solution is High-Availability. This doesn't work with multiple Warpgate instances working in HA since you're only updating one instance's cache when you update the parameters. @Eugeny I assume we care about HA setups right? And if so, how would you solve it? A periodic cache refresh? Or an invalidation protocol that is sent across the instances?

Maybe we can do something where Warpgate is told whether it's HA (A config field?), and then enable a periodic cache refresh only if it's in HA mode? Since 99% of warpgate deployments are single-instance I'm pretty sure due to the complexity of setting up HA, so having an always-enabled cache refresh seems inefficient, even if it is a lot better than the current read DB on every HTTP request approach.

LarsSven:
So the way I've tried to solve HA is the following:
- If the DB is not SQLite, we assume that the setup may be HA and we do a periodic (1 minute) sync of the cache.
- If the DB is SQLite, we assume that the setup is not HA, and we do not sync the cache.

I'm not sure this is the cleanest approach, would this work for you or would you prefer a different approach?

LarsSven:
The more I think about it, the more I don't like the assumption of sqlite for non-HA and others for HA. For example it makes sense to run postgres/mysql even for single instance deployments.

Would there be a cleaner alternative?

Eugeny:
I would not bother caching this one specific request (not doing it when it's not needed is a good improvement though)

There are already multiple DB roundtrips on each request (access checks, logging) and local database access is cheap enough anyway to not warrant building caches for individual queries like this one.

If there were multiple `Parameters` selects within a single query I would add a per-query cache (in `UnauthenticatedRequestContext` but right now there aren't

LarsSven:
That's quite fair, I'll close this PR and open another PR with just the not doing it when it's not needed.

Issue #2194, opened 2026-07-12
Option to open HTTP targets in a new browser tab


Chanta007:
Hey folks, first off thanks for Warpgate. We're running it as the SSO bastion in front of our dev boxes and it's been really solid.

One small bit of friction we keep running into with the target list. When you click an SSH target, the web terminal opens in a new tab, so the target list stays put behind it. But clicking an HTTP target navigates the current tab through the proxy, so the launcher page goes away. For folks who bounce between a couple of HTTP targets (in our case code-server and Forgejo) plus a terminal, they lose the target list and have to hit back or reload to get it again.

There is a workaround today: cmd or middle click the row to force a new tab. That works fine, but a lot of our users don't think to do it, and it feels like something the UI could just handle.

Would you be open to a small opt-in setting for this? A couple of shapes that could work:

- A per user preference toggle, something like "open HTTP targets in a new tab"
- Or a per target option in the target config

I deliberately didn't suggest changing the default click behavior, since that could surprise existing users. Keeping it opt-in felt like the safer path. I had a look through the open issues and didn't spot an existing request for this, so apologies if I missed one.

If this sounds reasonable, I'm happy to take a crack at the PR myself. Just point me at where you'd prefer the setting to live (user prefs vs target options) and I'll follow your conventions. Thanks again for the project.

PR #2195, opened 2026-07-12
Opt-in: open HTTP targets in a new tab

Chanta007:
## Description

Follow-up to #2194 with a working implementation so there's something concrete to look at.

Clicking an HTTP target navigates the current tab through the proxy, so the target list page is lost. SSH and desktop targets already open in a new tab (via `window.open` in `openWebSsh` / `openWebDesktop`). This adds an opt-in preference so HTTP targets can behave the same way.

What's in here:

- A new `openHttpInNewTab` preference in `gateway/lib/store.ts`, built on the existing `autosave` helper so it persists in localStorage. Off by default, so nothing changes unless a user turns it on.
- `TargetList.svelte` now checks that preference in `loadURL` and uses `window.open(url, '_blank')` when it's set, mirroring the existing SSH pattern. The anchor also gets `target="_blank"` in that case so middle-click and keyboard activation stay consistent.
- A small footer toggle (`HttpTargetTabSwitcher.svelte`) next to the theme switcher, following the `ThemeSwitcher` pattern, with a tooltip that spells out the current behavior.

I kept it opt-in on purpose rather than changing the default click behavior, since that felt like it could surprise existing users. Happy to move the toggle somewhere else, rename the setting, or turn it into a per-target option instead if you'd prefer that shape. Just let me know what fits the project best.

Checks run locally in `warpgate-web`: `biome check` clean, `npm run check` (svelte-check) 0 errors / 0 warnings, and `vite build` succeeds.

## AI Usage

* [ ] Fully vibe coded


Chanta007:
Did a local end-to-end check on this branch and wanted to share what I verified.

I built the fork with the repo's `docker/Dockerfile` and ran it. The running binary reports `v0.27.0-beta.1-1-g02ec0ac`, so it's this branch. I seeded an HTTP target through the admin API and confirmed it shows up in the user selector.

I then inspected the frontend bundles the running server actually serves:

- The `App` chunk contains the footer toggle component, including both tooltip strings and the toggle that flips the `openHttpInNewTab` store.
- The `TargetList` chunk contains the new `loadURL` branch, minified to `function Z(e){b(F)?window.open(e,`_blank`):location.href=e}`, plus the `target="_blank"` binding on the anchor. So the store-gated new-tab path is present and wired in the served build.

Static checks in `warpgate-web` are green: `biome check`, `npm run check` (svelte-check, 0 errors / 0 warnings), and `vite build`.

Full transparency: I didn't capture a screenshot of the two-tab behavior. My browser setup hit the self-signed cert interstitial on `localhost` and I couldn't click through it in an automated way. Happy to add screenshots if that's useful, or to reshape the approach (for example a per-target option instead of a global toggle) if that fits the project better. Just let me know.

Eugeny:
Thank you! @all-contributors add @Chanta007 for code

PR #2217, opened 2026-07-15
Fix embedded session menu clicks misclassified as drags

Hexalyse:
## Description

Fix the embedded session menu failing to open with normal mouse clicks. The problem is confirmed manually in Brave.

Reason: Chromiym can emit a zero-distance `mousemove` between `mousedown` and `mouseup`. The menu treated any `mousemove` as a drag, causing `mouseup` to skip opening the menu.

This change records the initial cursor position on `mousedown` and requires 5 pixels of movement before entering drag mode. This threshold can be lowered to 1px if needed, it will still prevent the problem which was happening even when the cursor does not move at all between the click and the mousemove event.

The issue and root cause were manually verified in Chrome DevTools:

- Normal clicks produced `mousedown → mousemove → mouseup` with identical coordinates, which prevented the menu from opening.
- Device-emulated taps (via dev tools mobile device emulation) opened the menu.

The production frontend build and Biome checks pass.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/warpgate-web/src/embed/EmbeddedUI.svelte
+++ b/warpgate-web/src/embed/EmbeddedUI.svelte
@@ -6,9 +6,10 @@
     let ready = false
     let menuVisible = false
     let dragging = false
+    const DRAG_THRESHOLD = 5
     let savedPosition = { x: 0.1, y: 0.8 }
     let position = { x: 0.1, y: 0.8 }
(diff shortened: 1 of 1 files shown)


Eugeny:
Thanks! @all-contributors add @Hexalyse for code

PR #2220, opened 2026-07-16
Avoid unnecessary set_target and set_user_info hot path work

LarsSven:
## Description
Kubernetes and HTTP targets make use of set_target and set_user_info. These functions run once per request, and are quite expensive (JSON serialization and DB call). In 99% of cases they are however simply an expensive no-op as the DB state is already equal to the request state.

This PR guards this such that set_Target and set_user_info only run when they will actually change the state, as we have access to both the session state and the request state in the context already. 

Furthermore, the order of operations of the code was a bit scary, as the memory state was committed before the DB state, meaning that a DB update failure would lead to a desync between DB state and memory state. This was somewhat fine in the previous implementation, as it would fix itself on the next request if it got desynced. But in the new approach, we are running less DB updates, so it won't fix itself as easily. So reorder the operations to first execute a succesful DB commit before committing the memory change.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/warpgate-common/src/config/target.rs
+++ b/warpgate-common/src/config/target.rs
@@ -27,7 +27,7 @@ impl Default for KubernetesTargetCertificateAuth {
     }
 }
 
-#[derive(Debug, Deserialize, Serialize, Clone, Object)]
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, Object)]
 pub struct TargetSSHOptions {
     pub host: String,
     #[serde(default = "_default_ssh_port")]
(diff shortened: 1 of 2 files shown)


Eugeny:
Thank you for all the effort you're putting into these performance improvements :)

LarsSven:
@Eugeny no problem at all :)

Quick selfish question though. I noticed I'm not yet in the list of contributors. I have been contributing regularly for some time. Was this perhaps accidentally forgotten or is there a particular contribution I have to make to get on it? :)

Thanks and sorry for the bother.

Eugeny:
Certainly an oversight :)

@all-contributors add @LarsSven for code

PR #2221, opened 2026-07-16
Significantly improve authorization performance

LarsSven:
## Description
Looking through the authorization code, I noticed that there is some very significant potential for performance improvements. Its main issue is that authorization work requires 5 DB queries, which needs to be performed on every request.

This PR makes quite a few core changes that would drastically drop the amount of work needed for authorization. The main changes are:
- Remove the global config-provider mutex and converted its API to immutable shared access, as it should be safe to do this without the mutex. This dominates a lot of the changes in this PR, even though I expect its impact on performance to be minimal.
- HTTP requests initially fetch the target, only keep its name, do some work, and then later the target is looked up again by-name. Instead keep the full target and reuse it later so we don't have to do the second DB query.
- Create a single role-assignment query to replace the 5 individual queries needed during authorization
- Add DB indexes for data needed commonly during authorization
- Use ID-based authorization (`authorize_target_by_id`) for targets where this is possible (HTTP, Kubernetes, web SSH, ticket requests etc) to avoid an unncessary DB query, while keeping the old name-based authorization (authorize_target) for SSH, MySQL and PostreSQL,.

This PR is currently very light on testing and benchmarking, as I am running at the end of my workday and would like to get this PR out. Some things have not even been tested beyond compile-checking yet. I am planning to perform some proper testing and benchmarking on this on Monday, but I would definitely not mind a review before then.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/warpgate-common-http/src/auth.rs
+++ b/warpgate-common-http/src/auth.rs
@@ -208,7 +208,7 @@ impl RequestAuthorization {
         }
     }
 
-    /// Returns a user ID if present in the authorization context.
+    /// Returns a user ID if present in the authorization context or nil UUID
     pub const fn user_id(&self) -> Uuid {
         match self {
             Self::Session(auth) => auth.user_id(),
--- a/warpgate-core/src/auth.rs
+++ b/warpgate-core/src/auth.rs
@@ -6,7 +6,7 @@ use crate::ConfigProvider;
 pub async fn validate_and_add_credential<C: ConfigProvider>(
     state: &mut AuthState,
     credential: &AuthCredential,
-    cp: &mut C,
+    cp: &C,
 ) -> Result<bool, WarpgateError> {
     let credential_valid = cp
         .validate_credential(&state.user_info().username, credential)
--- a/warpgate-core/src/auth_state_store.rs
+++ b/warpgate-core/src/auth_state_store.rs
@@ -157,7 +157,7 @@ impl AuthStateStore {
     /// and pass the result to [`AuthStateStore::create`], so that concurrent
     /// logins don't serialise on the store lock while doing database I/O.
     pub(crate) async fn resolve_user_and_policy(
-        config_provider: &Arc<Mutex<ConfigProviderEnum>>,
+        config_provider: &Arc<ConfigProviderEnum>,
         login_protection: &LoginProtectionService,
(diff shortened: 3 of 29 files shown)


LarsSven:
Also since the Rust code for the authorization query is fairly verbose, what I'm essentially doing is this:
```
SELECT 1
FROM user_roles ur
JOIN target_roles tr ON tr.role_id = ur.role_id
WHERE ur.user_id = ?
  AND tr.target_id = ?
  AND ur.revoked_at IS NULL
  AND (ur.expires_at IS NULL OR ur.expires_at > ?)
LIMIT 1;
```

Eugeny:
Nice - I've extracted the common part of the queries and added `authorized_target_ids`

LarsSven:
@Eugeny I benchmarked all of the optimizations again in Warpgate 0.27:

Before any optimizations:
```
Stress test complete
  workers:       20
  elapsed:       30.38s
  requests:      1459
  successes:     1459
  failures:      0
  throughput:    48.03/s
  bytes read:    1307264
  latency mean:  0.414s
  latency p50:   0.417s
  latency p95:   0.453s
  latency p99:   0.464s
  latency max:   0.479s
  status codes:
    200: 1459
```

Round 1 optimizations: Concurrent DB access, HTTP client reuse, avoid unnecessary show_session_menu checks:
```
Stress test complete
  workers:       20
  elapsed:       30.06s
  requests:      11855
  successes:     11855
  failures:      0
  throughput:    394.42/s
  bytes read:    10622080
  latency mean:  0.051s
  latency p50:   0.050s
  latency p95:   0.063s
  latency p99:   0.069s
  latency max:   0.094s
  status codes:
    200: 11855
```

Round 2 optimizations: Authorization rework, avoid unnecessary set_target/set_user_info:
```
Stress test complete
  workers:       20
  elapsed:       30.01s
  requests:      36427
  successes:     36427
  failures:      0
  throughput:    1213.76/s
  bytes read:    32638592
  latency mean:  0.016s
  latency p50:   0.016s
  latency p95:   0.020s
  latency p99:   0.022s
  latency max:   0.228s
  status codes:
    200: 36427
```

We are looking at over 20x performance improvement in a localhost HTTP stress test.

PR #2111, opened 2026-06-27
Terminate HTTP target WebSockets on logout

snvtac:
## Description

Fixes #2106.

HTTP target WebSockets now subscribe to the owning HTTP session before the upstream WebSocket handshake starts. When logout or another session removal path removes that session, the proxy aborts both WebSocket pump tasks so the active target connection is torn down instead of staying usable after logout.

This also keeps ticket-based HTTP WebSocket access working: ticket sessions without a SessionStore close receiver are allowed through, while normal user sessions still require a tracked session handle.

Added regression coverage for an HTTP target WebSocket that echoes successfully, logs out through `/@warpgate/api/auth/logout`, then observes the WebSocket closing.

Validation:

* `git diff --check`
* `rustfmt --edition 2024 --check warpgate-protocol-http/src/session.rs warpgate-protocol-http/src/proxy.rs`
* `python3 -m py_compile tests/test_http_websocket.py`
* `RUSTC_BOOTSTRAP=1 CARGO_HOME=$PWD/../cargo-home CARGO_TARGET_DIR=$PWD/../cargo-target cargo check -p warpgate-protocol-http --all-features`
* `RUSTC_BOOTSTRAP=1 CARGO_HOME=$PWD/../cargo-home CARGO_TARGET_DIR=$PWD/../cargo-target cargo test -p warpgate-protocol-http --all-features`
* `RUSTC_BOOTSTRAP=1 CARGO_HOME=$PWD/../cargo-home CARGO_TARGET_DIR=$PWD/../cargo-target cargo clippy -p warpgate-protocol-http --all-features --no-deps`
* `RUSTC_BOOTSTRAP=1 CARGO_HOME=$PWD/../cargo-home cargo build --all-features`
* local smoke: start Warpgate, open HTTP target WebSocket, logout, verify the active WebSocket closes

Note: `cargo clippy -p warpgate-protocol-http --all-features --no-deps -- -D warnings` still fails on an existing warning in `warpgate-protocol-http/src/api/sso_provider_list.rs:358` (`clippy::iter_kv_map`), outside this PR's changes.

## AI Usage

Choose the level of AI involvement for this PR.

* [ ] Fully vibe coded
* [x] AI-designed, AI-coded, manually checked
* [ ] Human-designed, AI-coded
* [ ] Human-designed, human-coded (includes AI autocompletions and boilerplate gen)

*<sub>This is not to block AI contributions but rather to speed up PR review (saves time on trying to deduce the logic behind AI hallucinations).</sub>*

PR #2234, opened 2026-07-20
Revert "Terminate HTTP target WebSockets on logout (#2111)"

LarsSven:
This reverts commit d8f898ab02c086b228bc46a342ea706c442b9cef.

## Description
I created an issue sometime ago in Warpgate about it not closing websockets on logout, which meant you could continue sending websocket traffic after you logged out. As a response to this, a contribution was opened that closed websockets when an HTTP session terminated, which got released on 0.26.

We use a lot of code-server (web-based vscode) through warpgate HTTP, and we started noticing vscode throwing WebSocket failures in the new version: `The workbench failed to connect to the server (Error: WebSocket close with status code 1006)`. This always happened after Warpgate terminated an HTTP session, and could only be fixed by clearing your browser cookies. Until you cleared your browser cookies, vscode would no longer be usable.

One thing I noticed is that Warpgate was closing the websockets forcefully, without sending a graceful close signal. This would have been a very logical reason for the vscode issue, especially because vscode reported a forceful websocket close. However, I tried to add graceful closure, but then during testing I would still get the issue in vscode, even when trying to gracefully close.

Furthermore, session handling in Warpgate is a bit weird where it doesn't always seem to be starting new sessions properly after a previous session has been expired, so doing websocket closure is in general a bit scary on http session termination with the current setup.

I would personally argue that the websocket issues that we are now getting are a lot more important than not properly closing websockets on logout, so I would suggest to revert this change now, so it can make the next release, and then we can more long term look into a more stable, deeply tested version of this that does not cause serious websocket issues.

## AI Usage

* [ ] Fully vibe coded
* [x] Human-designed, human-coded (includes AI autocompletions and boilerplate gen)

### DIFF
--- a/tests/test_http_websocket.py
+++ b/tests/test_http_websocket.py
@@ -1,12 +1,6 @@
 import ssl
-import time
-
 import requests
-from websocket import (
-    WebSocketConnectionClosedException,
-    WebSocketTimeoutException,
-    create_connection,
-)
+from websocket import create_connection
 from uuid import uuid4
 
 from .api_client import admin_client, sdk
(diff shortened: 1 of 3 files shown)


LarsSven:
@Eugeny if you're not happy with reverting it, we can also add it as a config option to toggle the feature, but for our setup we need to somehow get rid of it as we're running exams in a very high-stakes and websocket-heavy environment and this bug is causing us a lot of issues.

Imo reverting is the cleaner option as this option shouldn't be optional if it works as intended, so if we add a toggle for it we have to add an API change that we later either have to revert or keep around in an ugly way.

Eugeny:
I think I've found the actual issue - a pure websocket connection does not keep the session alive, so unless there are requests being made, the entire session will timeout depite websocket traffic and kill it - #2246 should fix it - would appreciate if you could give it a test!

LarsSven:
Hi @Eugeny,

Thanks for looking into it! You are indeed correct that websockets did not keep the session alive. That is something that I actually discovered a week ago and was gonna fix when I had more free time outside of my current work project.

It however unfortunately does not fix this problem. This problem isn't necessarily that sessions are getting killed in the middle of usage. The problem is that the following happens:
1. User starts the vscode HTTP target session
2. User at some point disconnects and leaves for the day
3. At some point, the session naturally terminates, vacuums and terminates websockets, which is expected because the user has been disconnected for some time.
4. The user comes back the next morning, and the vscode is now entirely broken and no longer wants to work until you clear your session cookies.

So it's not a problem of users getting crashes in the middle of their session, it's that once a session is cleaned up at an expected moment after the user has left, if they open the target again their vscode instance is entirely broken.

One thing I also know about the websocket code here is that it hard terminates the websocket session. You're supposed to send a "close message" (in this case for warpgate to both the user and the target), so that's another known bug with this implementation.

But the reason I am proposing to revert this is that I implemented the close message, but I still in my testing got the completely broken vscode after a session correctly terminates when I disconnect, so I don't know any other solution currently than to revert the code

Eugeny:
Could you please test your exact vscode setup with the nightly build? Besides fixing the session timeout it also fixes the redirect behavior for websocket requests with a timed-out session cookie

LarsSven:
I tested against 0.27.0-beta.2 and it indeed seems fixed. I think the redirect behaviour fix did it. Thanks :)

Eugeny:
Thanks for verifying 🤝

Issue #1445, opened 2025-07-31
HTTP Targets: injecting upstream HTTP Basic Auth credentials possible?


Christian-Little-Tava:
Context:
We have two different applications that are only reachable via Warpgate. Both of these applications are protected by HTTP Basic Authentication and have no way to turn that off. When users authenticate to Warpgate, they are still prompted with the basic auth popup (e.g., typing admin:admin), which causes friction and confusion.

Problem 1:
We’d like Warpgate to automatically supply the required Basic Auth credentials to the backend so that users don’t see a second login prompt after they’ve already authenticated to Warpgate. Ideally, Warpgate would allow configuring upstream HTTP Basic Auth credentials per target and inject the corresponding Authorization: header into proxied requests. This would eliminate the redundant login without weakening security (Warpgate is the only entry point to the internal app).

Problem 2:
We also have a different resource where we need two distinct access levels: admin and readonly. We want to avoid shared static credentials on the backend (because they get copied/shared easily), so we’d like a way for Warpgate to have two separate targets. In other words, we'd have a admin target and a readonly target for that resource, and it would auto log you into whichever role that target represents. 

Is this something that can currently be done? Or is it not possible?

Eugeny:
1. You can include the basic auth credentials directly in the target URL
2. Warpgate includes the authenticated user in an HTTP header, so if your target can cooperate it could read that, but otherwise there's no other authentication mechanisms for HTTP

PR #2300, opened 2026-07-30
feat(http): forward the SSO ID token to HTTP targets on request

ugoenyioha:
## Summary

Adds a `${WARPGATE_SSO_ID_TOKEN}` placeholder that an HTTP target's custom header value may contain. Warpgate substitutes the authenticated user's OIDC ID token, so an upstream can perform its **own per-user authorisation** rather than trusting a static shared credential.

Example target header:

```
Authorization: Bearer ${WARPGATE_SSO_ID_TOKEN}
```

## Motivation

Today an HTTP target has two options for upstream auth, both noted in #1445:

1. embed static credentials in the target URL, or
2. have the upstream trust `x-warpgate-username`

Option 1 means a shared secret that every user of that target effectively holds. Option 2 requires the upstream to trust a header, which only works if it can be sure the request came from Warpgate.

Forwarding the ID token instead lets the upstream verify a signed assertion of *who the user is*, and make its own decision — for example exchanging it with Vault/OpenBao JWT auth, so the secret store authorises per-user and its audit log names the human rather than a service account.

## Design notes

- **Opt-in by construction.** A target only receives the token if an operator explicitly puts the placeholder in a header value, so tokens are never leaked to unrelated upstreams. No new config field or schema change.
- **Fails closed on absence.** Sessions without an SSO token (password or ticket auth) have the header *dropped* rather than sent containing the literal placeholder, which an upstream would otherwise treat as a malformed credential. This is logged at debug.
- Applied to both the normal and websocket proxy paths.
- The token is never logged.

## Testing

`cargo check -p warpgate-protocol-http` is clean on top of `main`.

Verified manually against a live deployment: an SSO-authenticated session reaches the upstream with the header populated, the upstream successfully exchanges the token with OpenBao's JWT auth for a per-user token, and a password-authenticated session correctly arrives with the header absent.

Happy to adjust the placeholder syntax or move it behind an explicit `TargetHTTPOptions` field if you'd prefer that shape.

### DIFF
--- a/warpgate-protocol-http/src/proxy.rs
+++ b/warpgate-protocol-http/src/proxy.rs
@@ -165,13 +165,52 @@ fn copy_client_response<R: SomeResponse>(
     server_response.set_status(client_response.status());
 }
 
-fn rewrite_request<B: SomeRequestBuilder>(mut req: B, options: &TargetHTTPOptions) -> Result<B> {
+/// Placeholder that a target's custom header value may contain to receive the
+/// user's OIDC ID token. Opt-in by construction: a target only ever sees the
+/// token if an operator explicitly asks for it, so tokens are never leaked to
+/// unrelated upstreams.
+///
+/// Example header value: `Bearer ${WARPGATE_SSO_ID_TOKEN}`
+pub const SSO_ID_TOKEN_PLACEHOLDER: &str = "${WARPGATE_SSO_ID_TOKEN}";
+
+fn rewrite_request<B: SomeRequestBuilder>(
+    mut req: B,
+    options: &TargetHTTPOptions,
+    sso_id_token: Option<&str>,
+) -> Result<B> {
     if let Some(ref headers) = options.headers {
         for (k, v) in headers {
+            if v.contains(SSO_ID_TOKEN_PLACEHOLDER) {
+                // Drop the header entirely when there is no token (e.g. a
+                // password-authenticated or ticket session) rather than sending
+                // the literal placeholder, which the upstream would treat as a
+                // malformed credential.
+                let Some(token) = sso_id_token else {
+                    debug!(
+                        header = %k,
+                        "Skipping header: it requests the SSO ID token, but this session has none"
+                    );
+                    continue;
+                };
+                req = req.header(
+                    HeaderName::try_from(k)?,
+                    v.replace(SSO_ID_TOKEN_PLACEHOLDER, token),
+                );
+                continue;
+            }
             req = req.header(HeaderName::try_from(k)?, v);
         }
     }
     Ok(req)
+}
+
+/// The raw (compact JWT) OIDC ID token for this browser session, if the user
+/// authenticated via SSO. Kept out of logs deliberately.
+async fn sso_id_token_for(req: &Request) -> Option<String> {
+    let session = <&Session>::from_request_without_body(req).await.ok()?;
+    session
+        .get_sso_login_state()
+        .map(|state| state.token.to_string())
 }
 
 fn rewrite_response(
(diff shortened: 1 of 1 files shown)


FWest98:
This is a very interesting addition that we also welcome in our team! However, forwarding an ID token is technically not correct, since ID tokens are not intended to be used as such. See also https://auth0.com/blog/id-token-access-token-what-is-the-difference/.

In the scenario here, the user authenticates with warpgate, and warpgate then relays this authentication information to a backend application. In an OIDC scenario, WG gets two tokens: an ID token and an access token. The ID token has WG set as audience (`aud` claim), and is intended to be used only by warpgate. The access token, on the other hand, can be used by warpgate to pass to another application, such as the backend application.

Concretely, it would be best if this MR is adapted to support forwarding the _access token_. If some users really really want to use their ID token incorrectly, this could be supported as well, but I would strongly recommend supporting at least passing the access token.

ugoenyioha:
@FWest98 You're right, and thank you for catching it — I'm closing this rather than patching it, because I think the problem is deeper than the parameter it forwards.

I went and tested this properly after your comment, in a setup where a TLS-decrypting inspector sits between Warpgate and the backend application. Measured over one ~30 minute session against a real appliance:

- **420 crossings** carrying the ID token, all with `aud` = the Warpgate client_id — so the backend and the secret store were both consuming a token explicitly addressed to Warpgate. To make that work I had to configure the resource server's `bound_audiences` to accept Warpgate's client_id, which is the anti-pattern made load-bearing in config.
- The `sub` is an email address, and `email` is in the claims, so **PII crossed the inspected boundary 420 times**. AWS's guidance calls this out specifically as a reason not to send ID tokens to resource servers.
- The backend's SPA polls on a timer — **~1 request every 2 seconds from a single idle tab**, each carrying the token. So this isn't "a token is forwarded on an action"; it's a continuous credential stream for as long as a tab is open.
- ID tokens carry no `scope`, so there is no way to express what the forwarded credential may be used for. Authorization ends up entirely in the resource server's config rather than in the token.
- And as you note, there's no sender-constraining story: DPoP/mTLS binding apply to access tokens. I tried RFC 8693 token exchange to downscope before forwarding, and it's foreclosed from this position — the IdP (WSO2) requires an *access token* as `subject_token`, and Warpgate only exposes the ID token. So the mitigation for the anti-pattern is itself blocked by the anti-pattern.

OWASP ASVS 5.0 V10 puts it normatively: *"OIDC ID Tokens ... shall only be consumed by the client that triggered the authorization flow."*

I could add an access-token placeholder alongside this one, but I don't think that's the right shape either — it would leave the incorrect option in the codebase as a supported path, and "if some users really want to use their ID token incorrectly" is a footgun I'd rather not ship into a security product with my name on it.

What I think the correct version looks like, and what I'd like to come back with:

1. Forward the **access token** as the default and only sanctioned mechanism, with the audience for the target made explicit.
2. Where the IdP supports it, have Warpgate perform an **RFC 8693 exchange** and forward the *result* — a token with an audience bound to the specific target, and `act` set to Warpgate, so the backend can see both who the user is and that Warpgate is acting on their behalf. That's the delegation model RFC 8693 was designed for, and it gives the backend a real basis for authorization rather than "trust this token that wasn't meant for you".
3. Fall back cleanly, and loudly, where the IdP can't do the exchange.

Closing this so it can't be merged as-is. Happy to open a fresh PR along those lines if that direction sounds right to you — and if you'd rather it stayed a narrower change, I'd welcome the steer before I write it.

Thanks again for the review; the correction was the useful outcome here.

Eugeny:
:|

FWest98:
Well, that was not quite the intention. I would be strongly in favour of adding support for forwarding an access token, and indeed doing token exchange would be the "most correct" options here.

@Eugeny Maybe good to follow up on this separately?

Eugeny:
Yeah - I've set up a separate ticket to track ideas for what else we could put into the templating system

PR #2308, opened 2026-07-31
fix(http): don't forward Warpgate's session cookie to targets

ugoenyioha:
## The problem

Warpgate serves targets on the same origin as its own UI, so the browser attaches `warpgate-http-session` to every target request. The proxy then forwards the request verbatim — `copy_server_request` copies all headers except a name-based denylist, and `Cookie` isn't on it.

The result is that **every target, and anything able to observe the connection to one, receives a bearer token for the gateway itself**, on every request, valid for the cookie's full lifetime.

Nothing downstream wants it. In the deployment where I found this, the backend deletes it unread.

## Why this is a completion of existing design, not a correction

`may_forward_header` already withholds anything named `x-warpgate-*` from targets, on the principle that Warpgate's internals aren't the target's business. The session cookie is the same class of secret, but it travels under the name `Cookie`, so the name-based filter never sees it.

Warpgate has already authenticated the request from that cookie *before* proxying. Past that point it means nothing to any target.

## Why it matches by value, not by name

Adding `COOKIE` to `DONT_FORWARD_HEADERS` would drop **all** cookies and break every target that keeps its own session. So the filter has to be per-value.

But matching on the *name* `warpgate-http-session` has a real failure case: **a second Warpgate behind this one** uses the same cookie name with a different session id. Stripping by name would remove the inner gateway's cookie and break its login.

So the cookie is identified by **value**. `CookieConfig::default()` is `CookieSecurity::Plain`, so the cookie's value is the session id, which `SharedSessionStorage::load_session` already exposes. Consequences:

- a nested Warpgate's cookie has a different value → **survives**
- a target coincidentally using that cookie name → **survives**
- no resolvable session (ticket auth, stale cookie) → nothing matches → **safe no-op**

There's a test named for the nested case.

## Scope

- `copy_server_request` gains the parameter and a `COOKIE` branch; both the HTTP and WebSocket paths share it
- `session.rs` gains one accessor, `poem_session_id`
- 5 tests; each of 4 mutations (match by name only, never strip, always strip, strip when there's no session) is caught by at least one
- No configuration added — see below

## Measured

Deployed behind a TLS-terminating middlebox with a full header census (names only, values never recorded):

| | before | after |
|---|---|---|
| requests carrying `warpgate-http-session` | 1001 / 1001 | **0 / 365** |
| target's own cookies preserved | — | all 5, every request |

Plus a controlled check: 6 probes that deliberately sent the gateway cookie produced **0** cookie headers at the boundary.

## On configurability

I've deliberately not added a flag. An opt-in security fix protects nobody who doesn't already know about the problem, and an opt-out is a "forward my gateway session to every target" switch. With value-matching, the one breakage case I could construct is gone. Happy to add one if you'd prefer it.

## Not included

`copy_server_request` joins repeated header values with `"; "`, which is right for `Cookie` and wrong per RFC 9110 for most other headers. Pre-existing and unrelated — I've left it alone rather than fold it in, but can open a separate issue.

### DIFF
--- a/warpgate-protocol-http/src/proxy.rs
+++ b/warpgate-protocol-http/src/proxy.rs
@@ -29,8 +29,8 @@ use warpgate_tls::{TlsMode, configure_tls_connector};
 use warpgate_web::lookup_built_file;
 
 use crate::client_cache::HttpClientCache;
-use crate::common::SessionExt;
-use crate::session::SessionStore;
+use crate::common::{SESSION_COOKIE_NAME, SessionExt};
+use crate::session::{SessionStore, poem_session_id};
 
 static X_WARPGATE_USERNAME: HeaderName = HeaderName::from_static("x-warpgate-username");
 static X_WARPGATE_AUTHENTICATION_TYPE: HeaderName =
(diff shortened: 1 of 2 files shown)


Eugeny:
Sorry, too much slop. I've pushed my own fix in 0afa9703d38c25061295843d3e8aae2cb325f601

PR #2322, opened 2026-08-01
fix: don't send Transfer-Encoding: chunked on bodyless proxy requests

britbennett:
When proxying requests without a body (e.g. DELETE, GET), the proxy unconditionally wraps an empty stream with reqwest::Body::wrap_stream. This causes reqwest to add a Transfer-Encoding: chunked header to the outgoing request. Some upstream servers (notably Fastify) do not handle an empty chunked body gracefully and respond with 500 ISE.

## Description

...

## AI Usage

* [ ] Fully vibe coded
* [X] Human-designed, AI-coded


Eugeny:
Thank you! @all-contributors please add @britbennett for code

PR #2324, opened 2026-08-02
Implement reading targets from static config

### LINKED ISSUE #2276: [Feature Request] Allow targets and target groups to be specified in a static file
Hi,

I was wondering If It would be possible to specify the targets and target groups as a simple configuration file.

My use case is that I use Netbox to manage my entire infrastructure, and I can export (manually or via Netbox's API) pretty much any configuration file based on my hosts registered into Netbox. I would like to, periodically, fetch the available targets and target groups from Netbox as a static file (perhaps with a tag on an IP address, etc.) and reload Warpgate to make the targets immediately available.

Right now, It is possible but very tedious, as it needs to make multiple API calls to Warpgate to create target groups, targets, PATCH/PUT the target to modify settings, etc. All of It while handling the duplicate names, etc.

This would greatly improve users that chose Warpgate to access their entire infrastructure !

Cheers,

samtoxie:
## Description

Closes #2276.

Adds support for defining targets, target groups, and roles in a static YAML file that Warpgate keeps in sync with the database; so an external inventory (e.g. exported from Netbox) can be dropped in and periodically regenerated instead of driving the admin API by hand and reimplementing dedup/upsert logic.

### Config

```yaml
targets_file: targets.yaml # optional, resolved relative to config.yaml
```

Warpgate watches this file (same debounced `notify` pattern already used for `config.yaml`) and re-syncs on every change, plus once at startup.

### File format

```yaml
roles:
  - name: netbox-prod
    description: Auto-imported from Netbox

target_groups:
  - name: prod-web
    color: primary

targets:
  - name: web01.prod
    group: prod-web
    allow_roles: [netbox-prod]
    ssh:
      host: 10.0.0.5
      port: 22
      username: root
```

Any target kind (`ssh`/`http`/`mysql`/`postgres`/`vnc`/`rdp`/`kubernetes`) is supported, reusing the existing `TargetOptions` shape from the admin API.

### Reconciliation behavior

Each sync is a full reconciliation against the file's current contents:
- Rows are upserted by name and tagged `static_managed = true` (new DB column on `targets`, `target_groups`, `roles`).
- A previously-synced row whose name is no longer in the file gets deleted (with the same cascade cleanup (role assignments, tickets, known-host entries) as the admin API's own delete endpoints).
- A name that collides with an existing admin-created (`static_managed = false`) row is skipped with a warning, so a sync can never clobber something created by hand.
- The admin API now rejects (`409 Conflict`) edits/deletes of `static_managed` rows, since the next sync would just overwrite them anyway.
- When a target mentions a role or group that doesnt exist, the target is added but without that role or group. A warning is emitted in the logs to warn the operator of this.

### Admin UI

Targets, target groups, and roles that come from the static file are visibly and functionally read-only:
- List views show a "Static" badge (before the type label, for targets).
- Detail pages show an info banner explaining the row is managed by the static file instead of the API.
- Every form field is disabled including per-target-kind fields (SSH/HTTP/MySQL/Postgres/Kubernetes/VNC/RDP), role toggles, TLS settings, HTTP headers, and rate limits. This required threading a new `disabled` prop through several previously-non-disableable shared components (`TargetSshOptions`, `TargetVncOptions`, `TargetRdpOptions`, `TlsConfiguration`, `HttpHeadersEditor`, `RateLimitInput`).

### Known limitations

- Changing `targets_file` itself (adding/removing/renaming the path) requires a restart, only the file's *contents* hot-reload.
- Role assignments on a `static_managed` target are fully owned by the file: a role manually added to one via the admin API would be removed on the next sync.

### Testing

- New unit tests: static file YAML parsing, and an in-memory-sqlite reconciler test covering insert → update → delete and the admin-managed-name-collision case, plus config-key parsing tests.
- `cargo test --workspace` and `cargo clippy --workspace --all-targets` pass with no new warnings.
- `svelte-check` (0 errors/warnings) and `biome`/`eslint` pass on every touched frontend file (pre-existing unrelated lint issues elsewhere untouched).
- Regenerated `config-schema.json` and the admin OpenAPI schema/TS client to reflect the new config key and the `static_managed` field on `Target`/`TargetGroup`/`Role`.

## AI Usage

* [ ] Fully vibe coded


Eugeny:
Thank you for the PR! The approach LGTM but it will probably take me until the next week to get to reviewing this

samtoxie:
> Thank you for the PR! The approach LGTM but it will probably take me until the next week to get to reviewing this

Sounds good, no rush!

SteezyCougar:
This is awesome!

I think the biggest pieces I could see being helpful in addition to what you already have would be:
1. Env-var interpolation in the targets file. It would be nice to specify DB connections or whatnot, but obviously we'd want to pull in those creds from an env-var rather than hard-coded
2. User to role assignments!, and potentially even static user:target mappings?

samtoxie:
> This is awesome!
> 
> I think the biggest pieces I could see being helpful in addition to what you already have would be:
> 
>     1. Env-var interpolation in the targets file. It would be nice to specify DB connections or whatnot, but obviously we'd want to pull in those creds from an env-var rather than hard-coded
> 
>     2. User to role assignments!, and potentially even static user:target mappings?

Env interpolation indeed sounds like a good feature (although one could argue that a file configured with proper permissions is safer), if i find some time this weekend ill look into it.

As for the second one its interesting too, not really something i thought of myself as i delegate rbac and role mapping to my oidc provider. if there is enough demand i think its worth adding too

PR #2358, opened 2026-08-05
link straight to a prefilled ticket request

jwillmer:
## Description

Lets any tool that already knows which target a user needs hand them a link that opens the ticket request form with that target selected:

```
https://<warpgate>/@warpgate#/ticket-requests?target=prod-db-1&description=Investigating+ticket+4821&duration=4h
```

`description` and `duration` are optional. Any of the three opens the form; with none of them the page behaves exactly as it does today.

### The problem

The self-service request form (added in #1818) asks the user to pick a target from a dropdown. But in practice the user isn't choosing anything — something else already decided which machine they need, and they're copying a name across from another screen. The dropdown defaults to whatever target happens to be first in their list.

A ticket carries its own target, so picking the wrong one isn't caught anywhere. The ticket authenticates perfectly and drops the user on a different machine. Same bastion address, same host key, a valid credential, no error. The usual outcome is that the user assumes the tool that sent them there is broken, and the real cause — a mis-selected dropdown entry — is close to invisible.

That's a class of mistake worth designing out rather than detecting.

### Our use case

We run support software that helps our engineers work customer issues. When an engineer needs temporary access to a specific customer server, our tooling already knows exactly which host that is — it's the one attached to the support case.

Today it can only print instructions: *open Warpgate, go to ticket requests, find `vessel-gw-07` in the dropdown, request 4 hours.* Every step of that is a chance to land on the wrong machine, and the failure is silent when it happens.

With this change the tool hands over a link instead. The engineer clicks it, sees a form already describing the access they're about to request, and presses **Request ticket**. If the target requires admin approval, it still requires admin approval.

The same shape works for anything else that knows the answer in advance — access brokers, runbooks, chat bots, an internal wiki page with a per-server link.

### What it deliberately does not do

**The URL fills the form in. It never submits it.**

This is the rule the whole change is built around. A link is a piece of text that can arrive from anywhere — mail, chat, a page written by someone else — so it must be able to *propose* a request and never to *cause* one. With **Auto-approve when user already has role-based access** enabled, a link that submitted on load would mint a live credential from a click. The user still reads what's on screen and presses the button, exactly as before.

**No existing gate can be bypassed by a URL.** Allow users to request tickets, per-target disable, always-require-admin-approval, the duration cap, the use cap — all unchanged, all still enforced. This moves a value into a form field and nothing more.

**A target the user isn't allowed to request is named, not quietly swapped.** If the link points at something that doesn't exist, has ticket requests disabled, or simply isn't visible to that user, the form opens with nothing selected, a warning naming the target that was asked for, and the submit button disabled until they pick one themselves. Silently falling back to the first entry in the list would make a typo indistinguishable from success — which is the original problem again.

**A duration over the cap isn't silently clamped.** It prefills, the existing validator rejects it, and the submit button stays disabled so the user can see why. The server enforces the cap regardless.

**A description arriving from a URL is length-bounded**, to the same limit the API already enforces. An admin reads that text when deciding whether to approve, and it came from a link.

Unknown query parameters are ignored.

### Scope

Frontend only — one file, `warpgate-web/src/gateway/TicketRequests.svelte`. No API, schema, database or Rust changes, and no admin UI changes. Deep links already survive the login redirect, so a link sent to someone who isn't signed in yet still works after they log in.

## AI Usage

* [ ] Fully vibe coded

Designed from a written spec and reviewed by a human before submission.

### DIFF
--- a/warpgate-web/src/common/helpers.ts
+++ b/warpgate-web/src/common/helpers.ts
@@ -1,4 +1,14 @@
 import type { BootstrapThemeColor } from 'gateway/lib/api'
+import { router } from 'svelte-spa-router'
+
+/**
+ * Query parameters of the current hash route, e.g. `#/foo?a=b`.
+ * Read once at component init - svelte-spa-router keeps the component
+ * mounted when only the querystring changes.
+ */
+export function routeQueryParams(): URLSearchParams {
+    return new URLSearchParams(router.querystring ?? '')
+}
 
 export function getCSSColorFromThemeColor(color?: BootstrapThemeColor): string {
     // Handle capitalized color names from API (e.g., "Primary" -> "primary")
--- a/warpgate-web/src/gateway/ApiTokenManager.svelte
+++ b/warpgate-web/src/gateway/ApiTokenManager.svelte
@@ -5,10 +5,10 @@
     import { parseHumantimeDuration } from 'common/duration'
     import EmptyState from 'common/EmptyState.svelte'
     import { stringifyError } from 'common/errors'
+    import { routeQueryParams } from 'common/helpers'
     import Loadable from 'common/Loadable.svelte'
     import { api, type ExistingApiToken } from 'gateway/lib/api'
     import Fa from 'svelte-fa'
-    import { router } from 'svelte-spa-router'
     import CreateApiTokenModal from './CreateApiTokenModal.svelte'
 
     let tokens: ExistingApiToken[] = $state([])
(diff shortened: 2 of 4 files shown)


jwillmer:
Heads up on the red `biome` check — I don't think it's coming from this PR.

The same job reports `Checked 151 files. Found 14 errors.` on PRs that touch no frontend code at all, e.g. [this run](https://github.com/warp-tech/warpgate/actions/runs/30961349262) on `dependabot/cargo/version-bumps-1fde359491`. The count here is identical.

Twelve of those are `assist/source/organizeImports` on files unrelated to this change. `src/gateway/TicketRequests.svelte` does appear in the list, but it's flagged on `main` as well — the diagnostic is the `common/RelativeDate` / `common/AsyncButton` ordering near the top of the file, which this PR doesn't touch.

I left it as-is rather than putting an unrelated import shuffle into a feature diff, and fixing this one file wouldn't turn the check green anyway. Happy to open a separate PR running `biome check --write` across the tree if that'd be useful.

Eugeny:
Thanks! @all-contributors please add @jwillmer for code

Issue #221, opened 2022-07-30
Support U2F/FIDO


klotzambein:
OpenSSH supports hardware authentication through the two key types "ecdsa-sk" and "ed25519-sk". From what I can tell this is not supported by warpgate.

How hard would it be to implement this? If it is relatively straightforward I would be willing to create a PR in the following weeks.

Eugeny:
I've got no experience with U2F so I don't know how it works exactly, but SSH PK auth is handled by the [`server_read_auth_request_pk `](https://github.com/warp-tech/russh/blob/master/russh/src/server/encrypted.rs#L310) method here - you'd need to add support for decoding the key format (https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.u2f) and validating the signature.

I'm happy to take care of all warpgate-specific integration code if you can get the SK support into `russh`!

PR #2360, opened 2026-08-05
feat: Add WebAuthn/FIDO2/U2F passkey support

britbennett:
Users can register hardware security keys (YubiKey, Titan Key, any FIDO2 or U2F key) or platform authenticators (Touch ID, Windows Hello) as an authentication factor for logging into Warpgate.

## Description

The primary focus is **hardware key 2FA** (FIDO2/U2F keys as a second factor alongside passwords), but the implementation fully supports **passkeys** (platform authenticators like Touch ID, Windows Hello, iCloud Keychain) as well. Passkeys could theoretically serve as a password replacement (single-factor passwordless login), but that would require a broader decision about refactoring the login flow — see "Not included" below.

Once authenticated with Warpgate (including the passkey step), users access all downstream targets (SSH, HTTP, MySQL, PostgreSQL, Kubernetes, RDP, VNC) normally through their authorized session.

### Scope

- **What it protects:** Login to Warpgate's gateway web UI (the bastion itself)
- **What it supports:** FIDO2 keys, U2F keys (backward compatible), platform authenticators (Touch ID, Windows Hello, phone biometrics)
- **Cluster support:** Authentication endpoints are cluster-safe via `on_login_owner` — works correctly in multi-node deployments
- **Non-browser protocols:** When an SSH/MySQL/PostgreSQL credential policy requires a passkey, the existing web approval flow triggers a WebAuthn ceremony in the browser. The terminal session stays pending until the user touches their key in the approval page.
- **Credential policy:** Admins can require passkeys per-protocol (same as OTP today), e.g. `{"http": ["Password", "WebAuthn"], "ssh": ["Password", "WebAuthn"]}`

### What's included

**Backend:**
- `credentials_webauthn` database entity + migration (m00073) with user_id index
- `CredentialKind::WebAuthn` added to auth enums, credential policy, and auth state machine
- Registration ceremony endpoints (start/complete) using webauthn-rs SecurityKey API
- Authentication ceremony endpoints (start/complete) in auth.rs with cluster proxy support
- Admin API: list and delete credentials per-user (recovery path for lost keys)
- Self-service: register/delete own credentials from gateway profile
- Counter persistence after each authentication (cloned-key detection)
- Failed WebAuthn attempts recorded to existing brute-force protection system

**Frontend:**
- "Add passkey" button in admin user config and gateway profile page
- Registration modal matching existing credential UI patterns
- Login flow handles `WebAuthnNeeded` state with auto-start and graceful retry on cancel/timeout
- Auth policy toggles show "Passkey" option (appears after user registers one)
- Auto-policy upgrade on registration (sets Password + Passkey, same as OTP does)
- Out-of-band auth page triggers WebAuthn ceremony for non-browser protocol web approval
- "Add passkey" hidden when viewing other users (admin can't register a key for someone else — keys are hardware-bound)

**Tests:**
- 3 Rust unit tests: WebAuthn satisfies policy, multi-factor requires both, rejection leaves state unchanged
- 5 Python integration tests: endpoint access control, policy enforcement, replay protection, admin CRUD, regression safety
- All 237 existing tests pass unchanged

### Design decisions

| Decision | Rationale |
|----------|-----------|
| `webauthn-rs` crate (by Kanidm) | De facto Rust WebAuthn library, production-proven |
| SecurityKey API (not Passkey API) | Avoids CredProtect extension conflicts with older U2F keys while still supporting both platform and roaming authenticators |
| `userVerification: "discouraged"` | Touch-only, no PIN required — matches AWS IAM MFA UX |
| `attestation: "none"` | Maximum hardware compatibility, no manufacturer restrictions |
| `residentKey: "discouraged"` | Server-side credential storage, works with older YubiKeys that have limited resident key slots |
| Max 8 keys per user | Matches AWS IAM limit, allows primary + backup keys |
| RP ID from `http.external_host` | Consistent with existing config, includes port for non-standard setups |
| Challenge state in poem session | Single-use (consumed on complete), no DB persistence for incomplete ceremonies |
| `danger_set_user_presence_only_security_keys` | Disables CredProtect that causes "inconsistent parameters" errors on many authenticators |

### Security

- **Phishing-proof:** Challenge is cryptographically bound to the RP origin (WebAuthn spec guarantee)
- **Single-use challenges:** Session state consumed immediately on complete endpoint — prevents replay
- **Counter persistence:** Detects cloned authenticators (stale counter → rejected by webauthn-rs)
- **Credential isolation:** Keys loaded per-user, cross-user usage impossible
- **No secrets in logs:** Credential data, challenges, and attestation objects never logged
- **Rate limiting:** Failed WebAuthn auth attempts recorded to existing brute-force protection
- **TOCTOU-safe:** Credential count re-checked before insert (closes race between start/complete)
- **Label validation:** Empty or >255 char labels rejected
- **Constant-time verification:** Signature verification uses ed25519-dalek/p256 (constant-time internally)
- **Backward compatible:** Existing users/credentials unaffected, new table is purely additive

### Dependencies added

2 direct dependencies:
- `webauthn-rs = "0.5"` (features: `danger-allow-state-serialisation`, `danger-user-presence-only-security-keys`)
- `webauthn-rs-proto = "0.5"`

12 transitive dependencies (x509-parser, serde_cbor_2, base64urlsafedata, asn1-rs, etc.)

All MIT/Apache-2.0 licensed. `cargo deny check` passes clean. No known security advisories.

### Configuration

The `external_host` must match what users see in their browser's address bar. This is already a Warpgate requirement (for cookies, SSO, etc.) — WebAuthn uses the same value.

```yaml
# Behind a reverse proxy (typical production setup)
http:
  external_host: "warpgate.example.com"   # public domain (what users see)
  listen: "0.0.0.0:8888"                  # internal, proxied by nginx/caddy

# Direct access without proxy
http:
  external_host: "warpgate.example.com:8888"
```

No special WebAuthn configuration needed beyond what Warpgate already requires.

**DNS binding:** WebAuthn works with subdomain-based DNS binding (e.g. `target.wg.example.com`) because we enable `allow_subdomains` on the WebAuthn builder. The RP ID is derived from `external_host` (e.g. `wg.example.com`) and any subdomain of it is accepted as a valid origin. However, targets bound to completely unrelated domains (not subdomains of `external_host`) cannot use WebAuthn — this is a WebAuthn spec limitation that all implementations share.

**No `external_host` set:** Falls back to `localhost`. WebAuthn will only work when accessed via `https://localhost`. This is fine for development but in production, `external_host` should always be configured (it's already required for cookies and SSO to work).

### Reusable for future features

The WebAuthn verification logic is exposed as a reusable pub(crate) helper:

webauthn::verify_webauthn_authentication(services, session, &credential_json).await?;

This handles challenge consumption, signature verification, counter update, and last_used in a single call. Future features like step-up authentication (e.g. require key touch before issuing tickets, approving access, or performing destructive admin actions) can use this directly without reimplementing the ceremony logic.

---

### Not included:

**Note:** This PR adds WebAuthn support for Warpgate's **web/browser login** only. It does not address SSH `-sk` key types (`ecdsa-sk`/`ed25519-sk`) requested in #221 — that is a separate SSH-protocol concern. However, for non-browser protocols that support web approval (SSH, PostgreSQL), this PR integrates with the existing web approval flow so users can satisfy a WebAuthn policy requirement by touching their key in the browser approval page.

#### Suggested UX improvement (out of scope but would add after initial PR).

- **Passkey instead of password:** For a non-2FA, passkey only auth policy example (passkey substituting a password) the login UI forces a password. You can type an incorrect password to get past this and then submit passkey in next continue login. Perhaps initial login only takes username/ID then a password type authentication would be extracted into the continue login flow. This way a passkey instead of password could work.

- **Forced enrollment on first login:** When an admin sets a passkey requirement before the user has registered one, the user is currently locked out. A "grace period" flow — where the first login succeeds with password only but forces the user to register a passkey before accessing any targets — would enable admins to enforce passkey requirements from day one without coordination. This would benefit all credential types (OTP, passkey, etc.) and is a separate feature.

- **Resident key mode as a config option:** Currently residentKey is hardcoded to "discouraged" (server-side credential storage, no key slots consumed). This maximizes compatibility but means browsers can't auto-discover credentials. An admin-configurable option in warpgate.yaml could allow choosing between "discouraged" (current default — unlimited capacity, works with all keys, but some browsers require an extra click to select authenticator type) and "preferred" (stores credentials on the key — browsers auto-discover them for smoother UX, but consumes limited hardware key slots: 25-100 per YubiKey model). This would be a one-line config that maps to the WebAuthn residentKey parameter.
#### Bug fix (out of scope but would add after initial PR).

- **Pre-existing bug identified (follow-up PR):** When the last credential of a given type is deleted (e.g. the only OTP device, the only passkey), the auth policy editor's toggle disappears but the policy value still contains the credential kind. Clicking "Update" saves a policy requiring a credential the user can't satisfy, locking them out. This affects all credential types and is a broader fix that should be addressed separately to avoid conflating concerns.

---

## AI Usage

* [ ] Fully vibe coded
* [X] Human-designed, AI-coded

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,11 +194,27 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "asn1-rs"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5493c3bedbacf7fd7382c6346bbd66687d12bbaad3a89a2d2c303ee6cf20b048"
+dependencies = [
+ "asn1-rs-derive 0.5.1",
+ "asn1-rs-impl",
+ "displaydoc",
+ "nom",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 1.0.69",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b7f43a50ac4fdca5df8e885c21b835997f0a1cdee65494a6847694a98652d9d8"
 dependencies = [
- "asn1-rs-derive",
+ "asn1-rs-derive 0.6.0",
  "asn1-rs-impl",
  "displaydoc",
  "nom",
@@ -206,6 +222,18 @@ dependencies = [
  "rusticata-macros",
  "thiserror 2.0.20",
  "time",
+]
+
+[[package]]
+name = "asn1-rs-derive"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "965c2d33e53cb6b267e148a4cb0760bc01f4904c1cd4bb4002a085bb016d1490"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+ "synstructure",
 ]
 
 [[package]]
@@ -1024,6 +1052,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 
 [[package]]
+name = "base64urlsafedata"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b08e33815c87d8cadcddb1e74ac307368a3751fbe40c961538afa21a1899f21c"
+dependencies = [
+ "base64 0.21.7",
+ "pastey",
+ "serde",
+]
+
+[[package]]
 name = "bcder"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1332,9 +1371,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.6.6"
+version = "4.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "473c7e07f409a8d772161724aa8db6a765a2532a70f9667eeb7b49d3d02fbdca"
+checksum = "301b56658598e48f3648647ac6fc887be7e7108eddfa4e9b63fcf3ec58c0cadf"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -1342,9 +1381,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.6.6"
+version = "4.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7b48fea5a88e9ae728a2dcbedbfc0e730f7d60da42e1cb049a83c9fb8b789889"
+checksum = "94a65403d1a1bd28f7dc68eb8506e8874808ee5eecb59298de588e2e1407a078"
 dependencies = [
  "anstyle",
  "clap_lex",
@@ -1615,7 +1654,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e75b2483e97a5a7da73ac68a05b629f9c53cff58d8ed1c77866079e18b00dba5"
 dependencies = [
  "digest 0.10.7",
- "spin 0.10.1",
+ "spin 0.10.0",
 ]
 
 [[package]]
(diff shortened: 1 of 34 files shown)


Eugeny:
Thank you for the PR! Just for the sake of transparency, it will take make a few weeks to review based on current capacity

huguesTIS:
really good and useful feature
Question about a specific pattern we're planning: a shared "on-call" account where each rotation member registers their own key (up to the 8-credential cap), so anyone on duty can log in without a shared secret.
Does the audit trail capture which credential_id was used for a given login? From a read of the code I could only see last_used getting bumped, with audit events only on create/delete and no event on the auth itself. 
For a shared account, that's the one thing that would let us tell which team member actually connected. Might already be planned, or maybe I'm missing where it's tracked
happy to help test if useful

britbennett:
@huguesTIS I can certainly take a look at the logging. I'd like to pick your brain about your use case though. Would you plan on distributing the initial shared password to oncall? In its current state a password is UI required, but I propose changing this after this initial PR is merged (mentioned above). How do you imagine the hardware key provisioning flow for oncall? I talked about implementing a grace period flow above where admin can set passkey/hardware key auth policy requirement for 2FA and let the user add their passkey on initial login but if this is already satisfied/set from another oncall user the second user wouldn't be able to add their hardware key without having the other.

wc7086:
Thanks for working on this important feature. Since it changes the authentication path of a bastion/PAM, it may be worth requiring a careful human line-by-line security review before merging.

PR #2396, opened 2026-08-10
Allow admin to configure new tab preference

LarsSven:
## Description
https://github.com/warp-tech/warpgate/pull/2195 made HTTP targets consistent with other targets, where it can open them in a new tab. For our usecase we work with students. For homework assignments we love this feature, as the new tab is much less confusing for new students than the same-tab setup. However, we also run exams on warpgate, and exam hall browsers deliberately do not allow multiple tabs, so the entire application should be contained within a single tab.

Unfortunately this option can only be configured by the user, and not by the admin. For us, we would like it to be such that:
- For homeworks, students get new-tab by default, but are free to choose themselves
- For exams, students are forced to same-tab to not break the exam browser.

To implement this, I added a new global parameter  for opening targets in a new tab, with 4 options:
- Default on: User configurable but defaults to new-tab
- Default off: User configurable but defaults to same-tab
- Forced on: New-tab without the user being able to change this
- Forced off: Same-tab without the user being able to change this.

## AI Usage

* [ ] Fully vibe coded


LarsSven:
Hi @Eugeny,

Quick question, is this a feature you'd be willing to merge/support? If so, then I can already open the related Terraform MR while this is still in review :)

Eugeny:
In general I want to avoid a checkbox-fest of having an admin policy attached to everything but I can see how this is important in a kiosk mode :)

LarsSven:
That makes sense, thank you.

PR #2417, opened 2026-08-12
fixed #2392 - cast the log `values` JSON column to text when searching

### LINKED ISSUE #2392: Log search fails with 500 on PostgreSQL: lower(json) on the log.values column
## Description

`POST /@warpgate/admin/api/logs` returns 500 whenever `search` is non-empty. The query applies `lower()` to `log.values`, which is a `json` column, and PostgreSQL has no `lower(json)`.

Log search worked on 0.25.5 and broke after upgrading to 0.27.3. An empty `search` still works, because the filter is skipped.

## Environment

- Warpgate: 0.27.3
- Database: PostgreSQL 17.2
- Install: `ghcr.io/warp-tech/warpgate:0.27.3` on Kubernetes

## Reproduction

Search anything in the admin UI log view, or:

```
POST /@warpgate/admin/api/logs
{"limit": 100, "search": "test"}
```

## Logs

```
ERROR HTTP{session=... session_username=admin}: warpgate_common_http::logging: Request failed
method=POST url=/@warpgate/admin/api/logs
error=Error { source: Some(DatabaseError(Query(SqlxError(Database(PgDatabaseError {
  severity: Error, code: "42883",
  message: "function lower(json) does not exist",
  hint: Some("No function matches the given name and argument types. You might need to add explicit type casts."),
  position: Some(Original(275)),
  file: Some("parse_func.c"), line: Some(629), routine: Some("ParseFuncOrColumn")
}))))) }
```

## Minimal repro against the warpgate database

```sql
select count(*) from log where lower("values") like '%test%';
-- ERROR:  function lower(json) does not exist

select count(*) from log where lower("values"::text) like '%test%';
--  count
-- -------
--     24
```

## Cause

`warpgate-admin/src/api/logs.rs` passes the json column to `case_insensitive_search`:

```rust
q = q.filter(case_insensitive_search(
    search,
    [
        LogEntry::Column::Text,
        LogEntry::Column::Username,
        LogEntry::Column::Values,
    ],
));
```

`values` is `json` in the PostgreSQL schema. `lower()` takes `text`, and PostgreSQL does not cast `json` to `text` implicitly. SQLite accepts it because of its dynamic typing.

## Expected behavior

Log search returns results on PostgreSQL.

## Fix

Cast the column to text before `lower()`, or drop `Values` from the searched columns.

## Related

#1785 reports the same column failing with `operator does not exist: json ~~ text`.

huguesgr:
## Description

Fixes #2392.

`POST /@warpgate/admin/api/logs` returns a 500 for any non-empty `search` on Postgres:

```
error returned from database: function lower(json) does not exist
```

`case_insensitive_search` wraps each column in `lower()`, and `log.values` is a JSON
column. Postgres has no `lower(json)` overload and won't coerce implicitly — SQLite and
MySQL do, which is why this only surfaces on Postgres.

The `values` column is now cast to text before `lower()` is applied. `case_insensitive_search`
gains an expression-taking sibling so the other five call sites, which are all plain text
columns, are left alone. MySQL spells the cast target `char` rather than `text`, hence the
match on the backend.

`tests/test_api.py` already runs against Postgres, but its `get_logs` case passed
`search=""` — which short-circuits the filter entirely, so the broken path was never
exercised. It now passes a non-empty search.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -276,7 +276,11 @@ ADMIN_API_TEST_CASES: list[AdminApiTestCase] = [
     AdminApiTestCase(
         id="get_logs",
         permission=None,
-        call=lambda api, r: api.get_logs_with_http_info(sdk.GetLogsRequest(search="")),
+        # A non-empty search is what actually exercises the filter - an empty one
+        # is skipped, hiding e.g. Postgres rejecting lower() on the JSON column
+        call=lambda api, r: api.get_logs_with_http_info(
+            sdk.GetLogsRequest(search="test")
+        ),
         expected_statuses={200},
     ),
     AdminApiTestCase(
--- a/warpgate-admin/src/api/common.rs
+++ b/warpgate-admin/src/api/common.rs
@@ -1,6 +1,6 @@
 use sea_orm::prelude::Expr;
-use sea_orm::sea_query::{Func, IntoCondition};
-use sea_orm::{ColumnTrait, Condition, EntityTrait, ModelTrait, QueryFilter};
+use sea_orm::sea_query::{Alias, Func, IntoCondition, SimpleExpr};
+use sea_orm::{ColumnTrait, Condition, DbBackend, EntityTrait, ModelTrait, QueryFilter};
 use warpgate_common::{AdminPermission, AdminPermissionSet, WarpgateError};
 pub use warpgate_common_http::RequestAuthorization;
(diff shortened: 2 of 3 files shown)


Eugeny:
Thank you! @all-contributors add @huguesgr for code

PR #2397, opened 2026-08-10
Add SSH certificate authentication for targets, issued by Vault

janisdombr:
Warpgate authenticates to an SSH target with a short-lived OpenSSH user certificate signed on demand by HashiCorp Vault, instead of a private key it stores. The ephemeral keypair is generated per connection and never persisted, so a compromise of the Warpgate host yields nothing a target would accept.

Targets trust the CA through TrustedUserCAKeys and need no authorized_keys. The certificate's key ID carries the Warpgate username and session UUID, so the target's own sshd log attributes a proxied session to a person rather than to the gateway.

VaultAuth offers workload identity only — kubernetes, AppRole, AWS, Azure and GCP. Each reads its credential from a file or a metadata service, never from the config: a static Vault password would merely relocate the long-lived secret this feature exists to remove. Full compatibility with OpenBao is supported.

Verified end to end against real infrastructure — AWS STS, a GCE instance, an Azure VM and a k3d cluster.

tests/test_ssh_target_cert_auth.py runs against a stub issuer and needs neither Vault nor a cluster.

Discussion: https://github.com/warp-tech/warpgate/issues/26

Special thanks to @theredspoon for the detailed test, OpenBao evaluation, and security recommendations.

## Description

...

## AI Usage

Choose the level of AI involvement for this PR.

* [ ] Fully vibe coded
* [x] AI-designed, AI-coded, manually checked
* [x] Human-designed, AI-coded
* [ ] Human-designed, human-coded (includes AI autocompletions and boilerplate gen)

*<sub>This is not to block AI contributions but rather to speed up PR review (saves time on trying to deduce the logic behind AI hallucinations).</sub>*

PR #2427, opened 2026-08-14
Reject discovered IdP endpoints with a non-HTTP scheme

theredspoon:
## What

Reject any OIDC-discovered endpoint whose scheme isn't `http` or `https`.

## Why

`openidconnect` parses discovery endpoints as generic URLs with no scheme constraint, and two of them end up in a client-side navigation (`Login.svelte`/`AuthBar.svelte`: `location.href = ...`). A provider advertising `"authorization_endpoint": "javascript:..."` gets that assigned to `location.href` on the gateway's own origin.

This is input hygiene, not a closed vulnerability: `javascript:`/`data:` are never legitimate values here, so rejecting anything outside `http`/`https` is a three-line check with no behavior change for any real deployment.

## Where

Checked once in `discover_metadata()` (`warpgate-sso/src/metadata.rs`), the single choke point every consumer of discovery metadata goes through, before the document is cached. A rejected endpoint is a hard error (`SsoError::UnsupportedEndpointScheme`), matching how other discovery/config problems already fail in this code. `Login.svelte`/`AuthBar.svelte` also get a `navigateToExternalUrl()` helper as a second line of defense.

## Tests

New unit tests in `warpgate-sso/src/metadata.rs` cover valid `http`/`https` documents and rejection of `javascript:`/`data:`/`file:` on each affected endpoint. `cargo test -p warpgate-sso`, `clippy`, and `biome check` all pass.


Eugeny:
Is there an actual case where this can happen? A hostile OIDC provider is not really a part of the threat model - it doesn't need to bother with running javascript when it can just issues an admin login.


(no LLM replies please)

theredspoon:
Hey @Eugeny, human here.

What happened here is that once a security finding came up in one of my LLM reviews on #2397, I figured I would point a security sweep at the codebase and share whatever came up, however minimal in it. Better disclose than not. And to be honest, I wasn't investigating deeply, just leaning on quantity as its own quality.

Of course on the face of it if an OIDC provider is compromised, they can do a lot worse than a javascript injection.

I would consider this PR more hygiene than a gap-closer. I've updated the PR body accordingly.

Related to that, would you be open to me opening a PR to add more context to SECURITY.md to better assist LLMs? I wouldn't want anyone's time wasted with findings you wouldn't advance due to them falling outside of the declared scope. I think the following could help:

```
  - Privileged-by-design actors turning hostile are out of scope, regardless of how many tiers exist.
  - Hardening suggestions are welcome; their absence isn't itself a vulnerability unless independently reachable.
  - Malicious transitive dependencies are only in-scope with a confirmed exploit path, not just a bad advisory.
  - A trusted IdP's own authenticated users acting maliciously is between them and the IdP, not Warpgate.
  - Only code that shipped in a tagged release needs a formal advisory.
  - Judge severity against the most direct alternative attack the same actor already has.
```

Eugeny:
Yes - definitely! And sorry for the delay

theredspoon:
Opened #2509. Have a great day!

PR #2462, opened 2026-08-20
Stop Kubernetes listener from becoming unavailable on incomplete TCP handshake

LarsSven:
## Description
Warpgate's Kubernetes listener has the following properties:
1. All TCP handshakes are serially accepted on the connection accept loop
2. Warpgate does not enforce a timeout on how long a TCP handshake can take

As a consequence of this, when a client does not complete its TCP handshake, the Kubernetes listener becomes completely unavailable and the Kubernetes feature of Warpgate stops working entirely. We are dealing with issues where our Kubernetes listener just stops listening every few days, and we suspect it is because bots are hitting the port without finishing up their TCP handshake.

This is resolved by:
1. Handle TLS handshaking on the per-connection task
2. Add a 10s timeout for TCP handshaking

Most of this is implemented in the tls-listener dependency, which I pull in as part of this PR.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9285,6 +9285,7 @@ dependencies = [
  "md5",
  "poem",
  "poem-openapi",
+ "rcgen",
  "regex",
  "reqwest 0.13.4",
  "reqwest-websocket",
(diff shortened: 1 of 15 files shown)


LarsSven:
@Eugeny this is perhaps a security fix rather than a simple bugfix as this is a pretty easy to do DoS angle?

Eugeny:
I'll push a few changes here in a few hours

PR #2464, opened 2026-08-20
fix: sort SSH target menu with natural numeric order

### LINKED ISSUE #2433: [Bug] SSH target list sort
Hello,

When connecting to Warpgate via SSH, the list of target isn't sorted properly, for example, 210- is directly under 20- instead of being at the correct place

Image

I saw this issue (https://github.com/warp-tech/warpgate/issues/1632) fixed it for the UI, maybe it's using the old sort for this list?

BetterAndBetterII:
## Summary

The SSH target picker sorted labels with `String::cmp`, so names like `210-…` sat directly under `20-…`. The web UI already uses case-insensitive natural sort; this applies the same order in `TargetMenu`.

Fixes #2433

## Test plan

- [x] `cargo test -p warpgate-protocol-ssh --lib sorts_targets_in_natural_numeric_order`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4488,6 +4488,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "natord"
+version = "1.0.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Eugeny:
Thank you! 

@all-contributors add @BetterAndBetterII for code

PR #2468, opened 2026-08-21
fix(ssh): respect configured auth methods on rejection

fergusean:
When public-key authentication is rejected, the SSH server currently advertises every authentication method through `MethodSet::all()`, including methods disabled in global parameters.

Advertise the configured method set instead. This prevents public-key-only deployments from causing clients to fall back to password or keyboard-interactive authentication.

Tests: `cargo test -p warpgate-protocol-ssh`


Eugeny:
Thank you! @all-contributors add @fergusean for code

PR #2469, opened 2026-08-21
fix(ssh): preserve host key confirmation prompts

fergusean:
## Problem

When Warpgate paused its SSH connection-status display to ask a user to verify an upstream host key, an already-queued “connecting” status update could arrive afterward and paint over that prompt. The SSH session then looked stuck: the confirmation was still required, but no longer visible for the user to answer.

## Fix

Give connection-status updates a generation and discard updates queued before the display is paused, resumed, restarted, or completed. This preserves the interactive host-key confirmation while leaving host-key verification and SSH authentication behavior unchanged.

## Validation

All required GitHub Actions checks pass, including build, test, lint, formatting, schema, and security checks.


Eugeny:
Thank you! I'll continue this in #2478

PR #2470, opened 2026-08-21
fix: reject target rename when the name is already taken

### LINKED ISSUE #2339: Target name uniqueness is only checked when creating a target, not when updating one
Version: 0.27.1

Creating a target with a name that is already taken is rejected with API error: "Name already exists". Editing an existing target and changing its name to one that is already taken goes through without any complaint, so you end up with two targets under the same name.

# Steps to reproduce

1. Create a target named `example`.
2. Create a second target named `example2`.
3. Edit `example2`, rename it to `example`, save.
4. Both targets are now called `example`.

# Why it matters

Users address targets by name, user#target for MySQL and user:target for SSH. With two targets sharing a name it is not obvious which one a connection ends up on, and from the user's side the two are indistinguishable. On a gateway whose job is to control who reaches which machine, that seems worth catching.

# sqlite

There is no unique constraint underneath the API to catch this either. On SQLite:

```
sqlite> SELECT name, "unique" FROM pragma_index_list('targets');
idx_targets_name|0
sqlite_autoindex_targets_1|1
```

`idx_targets_name` is non-unique, and the only unique index is the primary key on id. So the check in the create path is the only thing enforcing this today.

I have only tested this with targets. Users and roles may or may not behave the same way, I did not check.

BetterAndBetterII:
## Summary
- Updating a target now returns 409 when the new name is already used by another target, matching create.
- Renaming a target to its own name still succeeds.

Fixes #2339

## Test plan
- cargo check -p warpgate-admin
- tests/test_api.py::test_update_target_rejects_duplicate_name


Eugeny:
Thanks!

BetterAndBetterII:
Thanks for merging.

PR #2496, opened 2026-08-27
Make self-service credential management config-seedable

sj-louw:
## Description

`allow_own_credential_management` previously could only be set post-install (no static config equivalent), forcing operators who want it disabled from day one to script an API call after every fresh install. Wire it into the existing config-migration-seed pattern already used by `ssh.host_key_verification/recordings.enable`, so `config.yaml` can set the initial value directly.

Separately, `/profile/api-tokens` had no permission check - any authenticated user could mint their own API token regardless of the `allow_own_credential_management` setting, unlike password/OTP/public-key/ certificate self-service which are already gated by it. Apply the same parameters_based_auth transform already used by the other credential kinds.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/warpgate-common/src/config/mod.rs
+++ b/warpgate-common/src/config/mod.rs
@@ -906,6 +906,13 @@ pub struct WarpgateConfigStore {
 
     #[serde(default)]
     pub log: LogConfig,
+
+    /// Seeds the `allow_own_credential_management` Parameters row on first boot (existing
+    /// installs are unaffected — this only sets the value when the row doesn't exist yet,
+    /// same seeding semantics as `ssh.host_key_verification`/`recordings.enable`). Unset
+    /// (`None`) preserves today's default of `true`.
+    #[serde(default)]
+    pub allow_own_credential_management: Option<bool>,
 }
 
 impl Default for WarpgateConfigStore {
@@ -923,6 +930,7 @@ impl Default for WarpgateConfigStore {
             vnc: <_>::default(),
             rdp: <_>::default(),
             log: <_>::default(),
+            allow_own_credential_management: None,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Eugeny:
I'm explicitly trying to get rid of these config file options to avoid the issue of config drift between cluster nodes - is there perhaps a betters solution here somewhere?

What does your setup look like - if you are automating a fresh setup with no users yet, why not just update the settings via API before you add the rest of the configuration via it anyway?

sj-louw:
@Eugeny Im deploying gitops style to kubernetes, using custom helm chart (mounting config as configmap on pods) trying to do config as much as possible in code to avoid manual setup via UI. Im also using the terraform provider.

Im not sure what your final goal is in terms of moving to, from config file, but a suggestion:

Would it it help if I change this and my other PR #2497 to rather make these values settable via environment variables? Essentially having default values for the vars that can be overwritten during deployment/run time. All nodes/pods in the cluster will get same values, so no drift.

Eugeny:
I see - since you're using Terraform, the provider already lets you control those parameters: https://registry.terraform.io/providers/warp-tech/warpgate/latest/docs/resources/parameters

An environment var is just about node-local source, same as a local config file, and has the exact same issue of node drift which is why I'm trying to avoid it in the first place.

My rough vision is to keep the cluster-wide configuration (including these parameters) only in the database as a single source of truth, with the future gitops route being Warpgate reading/pulling and reconciling a config centrally as a cluster instead of each node doing it individually against its own local file

Feel free to also post about your "ideal world" here (this discussion will inform the design of native gitops support: https://github.com/warp-tech/warpgate/discussions/2482)

sj-louw:
Ah thanks for https://registry.terraform.io/providers/warp-tech/warpgate/latest/docs/resources/parameters I missed that part. Let me work through those and see how far I get with settings params via terraform.

Closing this PR for now.

PR #2507, opened 2026-08-28
Align WebSocket header behaviour with HTTP behaviour

LarsSven:
## Description
We have a setup where we point multiple HTTP targets at a Kubernetes ingress. Kubernetes ingress does routing based on the Host header. To be able to point multiple Warpgate HTTP targets at the same ingress, but still route them to different internal services, we manually set the Host field to a different value through Warpgate's headers configuration for HTTP targets.

This works great for HTTP requests, however the WebSocket behaviour differs from HTTP behaviour. For HTTP, the manually set header overrides the automatic host header, however for WebSockets Warpgate simply adds the Host header twice, which then gets rejected by backend services.

This PR changes this behaviour for WebSockets, such that manually set headers overwrite the automatic headers, in the same way that it currently does so for HTTP.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/warpgate-protocol-http/src/proxy.rs
+++ b/warpgate-protocol-http/src/proxy.rs
@@ -64,6 +64,9 @@ trait SomeRequestBuilder {
     where
         HeaderValue: TryFrom<V>,
         <HeaderValue as TryFrom<V>>::Error: Into<http::Error>;
+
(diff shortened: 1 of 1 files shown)


Eugeny:
Makes total sense, thanks!

PR #2528, opened 2026-08-31
feat(web): session search/filter with improved UX

### LINKED ISSUE #2418: [Feature]  Ability to search/filters in the sessions page
The sessions page can get filled with lot of logs over time and it gets hard to find a session which happened long back. It would be nice to have a search and/or filters in the sessions page

ozanonurtek:
## Summary

Implements session search and filtering on the sessions page, addressing #2418.

### Changes

**Filter bar layout (Sessions.svelte)**
- Search input on its own full-width row
- Protocol, From date, and To date filters in a 3-column grid below the search
- Active only / Logged in only toggles on a third row with a **Clear filters** button that resets all filters in one click (only shown when any filter is active)

**Search input focus fix (ItemList.svelte)**
- Moved the search `<Input>` outside the `{#await $items}` reactive block so it is never destroyed/recreated when data reloads — fixes focus loss on every keystroke
- Added `hasItems` state variable tracked via the RxJS pipeline to control search input visibility without depending on the async block scope

### Backend compatibility
The backend query layer uses SeaORM's abstract query builder with no raw SQL — fully compatible with SQLite (default), PostgreSQL, and MySQL.

### AI disclosure
This PR was authored with AI assistance (OpenCode / Claude). The contributor reviewed all changes, verified the build and biome CI, and tested the UX fixes. Architectural decisions (keeping the query layer backend-agnostic, the ItemList reactive structure fix) were guided by the contributor.

### AI Usage

* [ ] Fully vibe coded

Closes #2418


Eugeny:
Thank you for the PR! For the sake of transparency, I'm currently low on time and it will take me a bit to review & merge this

Eugeny:
What's the point of a separate commands view vs. making it a filter in the audit log view?

ozanonurtek:
Hello @Eugeny 

Thanks for your interest and comment on this! 

They query different tables:

- session_commands: structured rows written by the SSH command detector during normal operation, each row has the command text, a FK to its target session, and a timestamp. The Commands page provides dedicated filters (command text, user, target, date range) and links each hit back to its session.

- log_entries: the general audit event stream (auth, role changes, session lifecycle, etc.). Commands don't appear here 
during normal operation, only as unstructured fallback text if the DB insert fails.

So the data isn't available through the existing log endpoint. If you prefer a different UI organization, e.g. surfacing commands as a tab within the log viewer, I am happy to restructure. 

Is this over-scoped/over-engineered? What do you think, as a user which approach would you prefer?

Issue #2520, opened 2026-08-29
A target that fails its SSH handshake can leave the connecting client waiting


janisdombr:
When a target's SSH handshake fails, the gateway detects it and closes its own session immediately — the log shows the refusal and `Closed session` within the same second. The connecting user's `ssh`, however, is sometimes not released, and waits until something kills it.

Reproduced on `eb38ff188` with no patches, using an ordinary public-key SSH target whose host answers with a version string it never terminates:

```python
import socket, threading

s = socket.socket()
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
s.bind(("127.0.0.1", 9999))
s.listen(8)

def serve():
    while True:
        c, _ = s.accept()
        try:
            while True:
                c.sendall(b"SSH-2.0-" + b"A" * 1024)
        except OSError:
            pass

threading.Thread(target=serve, daemon=True).start()
```

Point a target at `127.0.0.1:9999` and connect through Warpgate. Most attempts fail promptly and correctly; roughly one in a hundred leaves the client hanging.

The gateway's own log for a hanging attempt:

```
INFO   Connecting address=127.0.0.1:… username="root"
ERROR  Connection error error=Ssh(Version)
ERROR  Target connection failed error=invalid SSH version string
DEBUG  Event event=State(Disconnected)
INFO   Closed session
```

All of that lands within one second. The client was killed two minutes later.

### Frequency

Three occurrences in 271 untraced attempts, about 1.1%. Running `ssh` with `-v` or `-vvv` suppresses it entirely — 0 in 520 attempts — so it looks timing-sensitive on the client side.

### What did not fix it

Sending an explicit `handle.disconnect()` from `disconnect_server()`, in addition to closing the client's channels, on the theory that a session failing before any channel is open has nothing to close and therefore sends nothing. The hang still reproduced at the same rate (iteration 31 of 120). So that is not the whole mechanism.

### Not understood

A standalone reproduction — one gateway, one target, the same bad banner, 700 consecutive attempts — never reproduced it, while a suite that starts a fresh gateway for each attempt reproduces at the rate above. Whether that points at gateway startup state or at something else about the harness, I don't know.

### Not investigated

Whether the disconnect reaches the wire, and whether the inbound session's task is still alive at that point.

PR #2549, opened 2026-09-05
Let go of a client that stopped reading when its target dies (#2520)

janisdombr:
Both halves of #2520.

A dead target now tells the client the connection is over rather than only closing its channels: the channel writer gained a connection-level `Disconnect`, queued behind the closes so the ordering holds. And a teardown finds out when russh's loop has actually shut the stream down, through a `oneshot` rather than a sleep.

The harder case is a client that stops reading until its channel window is empty. Everything Warpgate would say to it goes through russh's bounded queue, which russh drains only while it has no data pending, and russh's own loop parks in a socket write to that same client. So the connection is ended from underneath: Warpgate owns the stream before handing it to `run_stream`, and the wrapper starts answering `ConnectionAborted` and wakes whichever direction is parked. No `unsafe`.

russh moves to 0.63.3, which enforces its inactivity timeout during stalled writes. That timeout is what releases the socket without this PR, at `ssh.inactivity_timeout + admin_approval_timeout + 10s` — 625s on defaults. An earlier version of this description said the socket was never released; that rested on one 700s run which does not reproduce, and it is withdrawn.

What the wrapper changes is when. `Closing the session due to inactivity` fires at ~17s and the session closes at ~22s throughout. Without the wrapper the socket then keeps the whole approval grace on top of that; with it, the socket goes with the session. Same tree, differing only by the wrapper's three files, russh 0.63.3 on both sides, released in seconds after the target dies:

| `ssh.inactivity_timeout` / `admin_approval_timeout` | without | with `transport_abort` |
|---|---|---|
| 15s / 30s — russh timer 55s | 52.67, 52.65, 52.52 | 22.95, 22.93, 23.04 |
| 15s / 120s — 145s | 142.72, 142.54 | — |
| 15s / default 600s — 625s | 627.9 | 22.96, 22.96, 23.32 |

`protocol_handler_dropped` lands with the session close in every run with the wrapper and at russh's timer in every run without it. A reading client is released in 2.7s either way, which is why that case does not discriminate. On russh 0.63.2 the wrapper released in 23.20s and 23.07s and its absence left the socket held at 120s, which is a window shorter than russh's timer and so says only that 0.63.2 does not release it early.

The approval term is in russh's timer so that a session waiting on an approval is not killed under it. Once the session has closed, that reason is gone and the grace is still spent — ten minutes and ten seconds on defaults, per stalled session. That is the whole of what this PR now claims.

Tests: 33 unit in the SSH crate, 6 of them on the wrapper, and 37 integration across the 13 SSH suites. Each wrapper test was checked by deleting the guard it covers and confirming it goes red; one earlier version passed 200 runs out of 200 with the whole wake path removed, and was repaired.

The stalled-client harness needs Docker and about a minute per iteration, so it is not in CI. It takes the approval timeout as a parameter and prints every parameter before its first number, and runs on Linux as well as macOS: https://github.com/janisdombr/warpgate/blob/bc8d6df82/tests/test_stall_unread_client_release.py

Fork CI and CodeRabbit: https://github.com/janisdombr/warpgate/pull/10

## Description

...

## AI Usage

* [ ] Fully vibe coded


Eugeny:
Thanks! I've moved the bulk of the fix upstream into russh and cleaned up the PR.

janisdombr:
Heads up that the PR changed after your comment, so the description is worth a fresh look.

I re-measured with the crate version held on each side: the stalled client is never released without the stream wrapper, on either 0.63.2 or 0.63.3, and still held at 700s — past where russh's own timer lands on the default approval timeout. So the wrapper is the variable, not the crate. The head restores it on top of your commit and keeps 0.63.3.

Happy to drop it again if you would rather solve this inside russh. I would just want to know what to measure.

Eugeny:
russh's timer is `ssh.inactivity_timeout` + `admin_approval_timeout` + 10, and the approval default = 10 min, so with WG_INACTIVITY=15s it fires at 625s. The 240s test run is below that, and I don't know what your 700s run params

Could you do one run without the wrapper on 0.63.3 with the admin_approval_timeout_seconds set to 30, WG_INACTIVITY=15s, OBSERVE_S=180, and RUST_LOG=debug? This will set russh timer to 55s. 

if the FD is still held at 180s, it's a russh bug

janisdombr:
You were right to push on this. Rebuilt and re-measured, and the PR's headline claim does not survive.

Your configuration — no wrapper, 0.63.3, `admin_approval_timeout_seconds=30`, `ssh.inactivity_timeout=15s`, russh's timer at 55s — releases the socket at 52.7s, three runs out of three:

```
release_s = 52.67, 52.65, 52.52      (observed to 180s)
```

with the reason in the log you asked for:

```
DEBUG log: timeout while writing  log.target="russh"  lib_inner.rs:541
DEBUG warpgate_protocol_ssh::server::russh_handler: Dropped
ERROR warpgate_protocol_ssh::server: Session failed  error=Inactivity timeout
```

It tracks the timer:

| `inactivity + approval + 10` | released at |
|---|---|
| 55s | 52.67, 52.65, 52.52 |
| 145s | 142.72, 142.54 |
| 625s | 627.9 |

The bottom row is the 700s run you asked about: `WG_INACTIVITY=15s`, approval left at its default. It releases at 627.9s. My earlier run in that configuration reported the socket still held at 700s, and that is the number the table's "never" rested on. **It does not reproduce, and I'm withdrawing it.** I can't tell you what it was — that build tree is gone — and the likeliest answer is that I labelled the log from `Cargo.lock` rather than from the binary that ran. The 240s runs were below the timer and never discriminated either. So russh does release it, and 0.63.3 does what you moved into it.

What is left is narrower. The session ends long before the socket does:

```
18:18:27 INFO   Closing the session due to inactivity
18:18:32 INFO   client: Disconnect
18:19:01 DEBUG  russh_handler: Dropped      <- 29s later = admin_approval_timeout + 10
```

The approval term is in the timer so a session waiting on an approval is not killed under it. Once the session has closed that reason is gone, but the whole grace is still spent — ten minutes and ten seconds on defaults, per stalled session. With the wrapper, same tree, differing only by its three files:

| | released at | `protocol_handler_dropped` |
|---|---|---|
| wrapper, approval 30s | 22.95, 22.93, 23.04 | with the session close |
| wrapper, approval default | 22.96, 22.96, 23.32 | with the session close |
| none, approval 30s | 52.67, 52.65, 52.52 | at russh's timer |
| none, approval default | 627.9 | at russh's timer |

So what the wrapper changes is that the socket goes with the session rather than with the approval grace — not that it goes at all. Whether that is worth 432 lines is your call and I would understand closing this. If it is worth something inside russh instead, the shape is that a session which has already closed should not still carry the approval term in its inactivity deadline.

The harness now takes the approval timeout as a parameter and prints every parameter before its first number, so any row above is a three-minute run instead of an eleven-minute one: https://github.com/janisdombr/warpgate/blob/bc8d6df82/tests/test_stall_unread_client_release.py

PR #2558, opened 2026-09-06
Audit Kubernetes exec, attach, port-forward and debug containers

huguesgr:
## Description

Warpgate proxies and records Kubernetes traffic, but emits no structured audit
events for it. A `kubectl exec` leaves only a free-text `"Request"` row with a
raw URL, and `kubectl debug` leaves even less: the image, command and target
container it asks for live in the request *body*, not the path, so they reach
the API recording only, and only when recording is enabled. Nothing in
Admin → Logs distinguishes `kubectl get pods` from someone attaching a
privileged ephemeral container to a running pod.

This adds six audit events, emitted whether or not session recording is on:

| `_type` | Fields (beyond session / user / target) |
|---|---|
| `KubernetesExecStarted1` | namespace, pod, container, command, tty, stdin |
| `KubernetesAttachStarted1` | namespace, pod, container, tty |
| `KubernetesPortForwardStarted1` | namespace, pod, ports |
| `KubernetesStreamRejected1` | namespace, pod, subresource, status |
| `KubernetesDebugContainerCreated1` | debug container, image, target container, command, tty, response status |
| `KubernetesPodCreated1` | images, node name, hostPID / hostNetwork / privileged, response status |

`KubernetesPodCreated1` deliberately audits *any* pod creation rather than
guessing at kubectl's intent from a `node-debugger-*` name — it covers
`kubectl debug node/...` and `--copy-to` without a heuristic, and the flags make
a debug pod obvious.

### Design notes

- A single classifier (`warpgate-protocol-kubernetes/src/audit.rs`) feeds both
  the audit events and the recording metadata, so the two cannot drift.
- It reads the **request** path rather than the upstream URL, so a target whose
  cluster URL carries a path prefix is still audited.
- Container specs are summarised to a fixed set of fields. `env`, `envFrom` and
  `volumeMounts` are never parsed — they routinely carry secrets — and argv is
  capped by both element count and total bytes so a client cannot make a log row
  unbounded. There's a test asserting no environment value can reach an event.
- Events carry `session` explicitly (as `UserAuthenticated1` already does),
  which is what lets the websocket upgrade closure emit at all: it runs outside
  the instrumented request span, and the log layer drops events with no session
  id.
- No OpenAPI regeneration needed — `LogEntry.values` is a free-form JSON map.

### Two existing bugs fixed along the way

- **The recorded exec command was wrong.** `kubectl exec pod -- ls -la` sends
  one `command=` query parameter per argv element, and `recording.rs` collected
  the query into a `HashMap`, so only one survived — recording metadata showed
  `-la` instead of `ls -la`. Metadata now holds the full argv, and an absent
  container is `None` rather than the literal string `"unknown"`.
- **Websocket failures never reached the audit log.** The upgrade closure runs
  outside the request span and the log layer drops session-less events, so an
  exec the cluster rejects left nothing in the session log. They now name the
  session — and a close-related failure, which is how every `kubectl exec` and
  `port-forward` ends, logs at `debug` instead of reading as an error.

### Verification

- 21 unit tests in `audit.rs` / `handlers.rs`, including real `kubectl debug`
  PATCH and PUT-fallback bodies, the repeated-`command=` regression, the
  redaction assertion, and the truncation budget.
- Three e2e tests in `tests/test_kubernetes_integration.py` covering exec,
  debug and port-forward, driving real `kubectl` against the k3s fixture.
- Manually exercised end to end against a real k3s cluster through a Warpgate
  instance: `exec`, `attach`, `port-forward`, `debug --target=...` and
  `debug node/...` all produce the expected rows in Admin → Logs.

One protocol subtlety worth flagging for review: the **websocket**
port-forward protocol negotiates ports per stream and sends no query string at
all, so `ports` is `Option` and simply absent there — an empty list would read
as "forwarded no ports", which isn't the same thing. Only the SPDY variant puts
`ports=` in the URL.

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9313,7 +9313,6 @@ dependencies = [
  "poem",
  "poem-openapi",
  "rcgen",
- "regex",
  "reqwest 0.13.4",
  "reqwest-websocket",
  "rustls 0.23.43",
--- a/tests/test_kubernetes_integration.py
+++ b/tests/test_kubernetes_integration.py
@@ -1,5 +1,7 @@
 from datetime import datetime, timezone, timedelta
 import base64
+import json
+import os
 import hashlib
 import html
 import re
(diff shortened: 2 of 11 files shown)


Eugeny:
Thanks!

PR #2578, opened 2026-09-13
feat(auth): support multiple authentication combinations per protocol with matrix UI

mtthsbrd:
## Description

Currently, protocol-level credential requirements only allow either "Anything" or a single conjunction (ANDed set) of credential requirements (e.g. Password AND TOTP). There was no way to specify alternative authentication pathways, such as allowing a user to authenticate via SSH using either **SSH Key** OR **Password + TOTP** OR **Web In-Browser Approval**. The previous UI also did not make credential combinations clear.

This PR introduces support for **multiple authentication combinations per protocol** (Disjunctive Normal Form: OR-of-ANDs), while preserving the "Allow any" option, paired with an interactive matrix UI in the web admin.

auth_policy_matrix

### Key Changes

- **Core & Data Models**:
  - Updated `UserRequireCredentialsPolicy` protocol fields (`ssh`, `http`, `vnc`, `postgres`, `kubernetes`, `rdp`) from `Option<Vec<CredentialKind>>` to `Option<Vec<Vec<CredentialKind>>>`.
  - Maintained backward compatibility in deserialization/migration so existing single-conjunction configurations seamlessly deserialize into the new format.
  - Updated session authentication pipelines across all protocols to validate whether the user's provided credentials satisfy any of the configured combination rules.

- **Web Admin (Matrix UI)**:
  - Added an interactive matrix editor in `AuthPolicyEditor.svelte`:
    - Rows represent alternative combinations (OR paths).
    - Columns represent credential kinds (AND within each combination).
    - Clear controls to add/remove combination rows and toggle credential requirements.
  - Modernized event handling and typing in `CredentialManager.svelte` and user management views.

- **API & Testing**:
  - Regenerated OpenAPI schemas, TypeScript types, and Python SDK.
  - Adjusted `oasdiff` severity rule for the intentional schema evolution.
  - Updated all 15 integration test suites across SSH, VNC, HTTP, Postgres, Kubernetes, and RDP to the 2D policy structure.

---

## AI Usage

Choose the level of AI involvement for this PR.

* [ ] Fully vibe coded
* [ ] AI-designed, AI-coded, manually checked
* [x] Human-designed, AI-coded
* [ ] Human-designed, human-coded (includes AI autocompletions and boilerplate gen)

*<sub>This is not to block AI contributions but rather to speed up PR review (saves time on trying to deduce the logic behind AI hallucinations).</sub>*

PR #2580, opened 2026-09-13
Make a failed wait say why it failed

janisdombr:
`_wait_timeout` raised only `if t.is_alive()`. A waiter thread that died from an exception is not alive either, so a failure that raised rather than running out the clock was reported to the caller as a port that came up.

Two waits rode on that:

- `wait_port`'s `for_process` branch raised inside a `try` whose `except` catches `TimeoutExpired`, so the one mechanism written to say "the process died" was swallowed by the line below it. A child that exits immediately returned success in 0.01s.
- `recv=True` raises a plain `Exception`, which is not a `socket.error`, so an open port that closes on us never reached the caller either.

`wait_mysql_port` had its own copy of the same thread block and now shares the fixed one.

Today `minio/minio` stopped being publicly pullable from Docker Hub, so `docker run` in `start_minio` exits at once. `main` reports `Port 1903 is not up` after 60s; `docker: pull access denied` is in pytest's captured stderr, which is printed only on failure and names no port. With this, the same run says which process exited and with what code.

`tests/test_util_wait_port.py` covers the two repaired paths, the plain timeout, and a port that is genuinely up. Against the current `tests/util.py` the two repaired ones fail and the other two pass.

The registry side is handled by #2581, now merged, which switches `start_minio` to `quay.io/minio/minio`. This PR is only about the wait reporting what happened. (An earlier version of this paragraph said #2578; that was a wrong number on my part — #2578 is an unrelated authentication PR and never touched the registry.)

## Description

...

## AI Usage

* [ ] Fully vibe coded

### DIFF
--- a/tests/test_util_wait_port.py
+++ b/tests/test_util_wait_port.py
@@ -1,0 +1,76 @@
+"""What `wait_port` tells the caller when the wait does not succeed.
+
+Every case here failed silently before: the caller was told the port was up.
+On 2026-09-13 that turned `minio/minio` disappearing from Docker Hub into
+`Port 1903 is not up` after a 60s timeout, with `docker: pull access denied`
+visible only in pytest's captured stderr.
+"""
+
+import socket
+import subprocess
+import threading
+import time
+
+import pytest
+
+from .util import alloc_port, wait_port
+
+
+def test_a_child_that_exits_is_named_rather_than_timed_out():
+    """The case the `for_process` argument exists for.
+
+    It raised inside a `try` whose `except` caught `TimeoutExpired`, so the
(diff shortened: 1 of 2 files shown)


mtthsbrd:
Did you want to reference https://github.com/warp-tech/warpgate/pull/2581 instead of https://github.com/warp-tech/warpgate/pull/2578 ?

janisdombr:
> Did you want to reference #2581 instead of #2578 ?

Yes — #2581, thanks for catching it. Fixed in the description.
#2578 was a wrong number on my part, and not a harmless one: I skipped the registry fix for a while because I believed someone else was already making it. #2581 is mine, opened once I checked.

PR #2584, opened 2026-09-14
Let the client read the reason before ending the session

janisdombr:
A session closed for inactivity does not tell the client why. The client sees the target's output, then nothing.

`disconnect_server` queues the connection-level `Disconnect` behind the channel closes, so it reaches the client in the same breath as whatever the session just said. A client handed both in one read acts on the disconnect and exits without printing what it already holds. The inactivity arm of the event loop emits its notice and calls `disconnect_server()` in the next statement, so that path loses its notice every time.

The bytes do reach the wire — russh appends the data, the close and the disconnect in that order and flushes before shutting the stream down, and a passing and a failing run are identical on the gateway side.

The disconnect now goes out from the task that already waits out the flush grace before closing the socket, and only to a client that is reading; one that is not still gets cut immediately. The channel closes stay in the queue, so per-channel ordering is unchanged, and the grace is not lengthened — `flushed` and `delay` are read, not altered.

New test, `tests/test_ssh_idle_close_message.py`. Ten runs per cell, rebuilt between cells:

| | failures out of 10 |
|---|---|
| without this change | 10 |
| with it | 0 |

`tests/test_ssh_proto.py::test_connection_error` is repaired in the same commit: it asserted `returncode != 0` without waiting for the client, and `"user:ssh-bad-domain@localhost"` sat where `ssh` takes the remote command rather than the destination, so the session reached the healthy target and succeeded. Its target now points at a port nothing listens on and the assertion names the message. It fails against the unpatched gateway too.

`cargo cranky --workspace --all-features` clean, 28 unit tests in the SSH crate, all 13 SSH integration suites green. The socket release from #2520 is unaffected: a stalled client is still let go at session close (22.9s, `protocol_handler_dropped` landing with `session_closed`) rather than at russh's timer, with a reading client released in 8.0s either way.

## Description

...

## AI Usage

* [ ] Fully vibe coded


janisdombr:
Folded into #2548, whose test needs it.

